PR #2, opened 2020-09-11
Upgrade Quarkus to 1.8.0.Final

rsynek:
**referenced Pull Requests**:

https://github.com/kiegroup/optaplanner/pull/922

### DIFF
--- a/quarkus-school-timetabling/build.gradle
+++ b/quarkus-school-timetabling/build.gradle
@@ -15,7 +15,7 @@
  */
 
 plugins {
-    id "io.quarkus" version "1.7.0.Final"
+    id "io.quarkus" version "1.8.0.Final"
     id "java"
 }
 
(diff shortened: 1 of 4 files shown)


quarkus-school-timetabling/build.gradle
```diff
@@ -65,3 +65,12 @@ java {
 test {
     systemProperty "java.util.logging.manager", "org.jboss.logmanager.LogManager"
 }
+
+repositories {
+    maven {
+        url "https://repository.jboss.org/nexus/content/groups/public/"
+        mavenContent {
+            snapshotsOnly()
```
ge0ffrey, line 73: Why does the gradle build needs this and the maven builds don't?

Ideally when we do a release, the "stable" branch doesn't have unnecessary code like this, as we expect many users to start from these quickstarts.

rsynek: Maven builds do need it too, see https://github.com/kiegroup/optaplanner-quickstarts/blob/development/quarkus-school-timetabling/pom.xml#L182.

The reason is that now we depend on kogito snapshots.

ge0ffrey: Argh, we'll need to delete this part everytime we sync the development branch to the stable branch (= default branch).

rsynek: That's one more step for the release pipeline, I guess.

PR #16, opened 2020-10-15
Upgrade to quarkus1.9.0

ge0ffrey:
Related: https://github.com/kiegroup/optaplanner/pull/974


ge0ffrey:
Ok this is wierd, this is only suppose to be 2 commits. Investigating what I did wrong gitwise (probably wrong branch started from).

Update: this is fixed

PR #17, opened 2020-10-18
Factorio layout quickstart

ge0ffrey:


quarkus-factorio-layout/src/main/java/org/acme/factoriolayout/domain/Area.java
```diff
@@ -0,0 +1,45 @@
...
+public class Area {
+
+    private int x;
+    private int y;
+
+    // No-arg constructor required for Jackson
+    public Area() {
+    }
+
+    public Area(int x, int y) {
```
yurloc, line 30: Add `@JsonCreator` on this constructor to get rid of the no-arg one. That will also allow you to make the class immutable.

ge0ffrey: Interesting to know, thanks for sharing.
 I looked into this, and for Area it works, but for Recipe, Requirement, etc. It doesn't work:
```
Invalid type definition for type `org.acme.factoriolayout.domain.RecipeInput`: Argument #0 has no property name, is not Injectable: can not use as Creator
```
This makes me believe it's unreliable: how does it know that x is the first parameter and y is the second? It could actually swap them around.

The official docs state that if we use this we have to use @JsonProperty() too, which creates more boilerplate which defaults the purpose of using it - or mess with the build (-parameters) which is not an option as it's too unvanilla.

Until JDK by default doesn't forget the parameter names in the class files, I'd rather not use this.

triceo (commented):
Personally, I don't think we need this. It doesn't showcase any new features, it doesn't integrate with any new technology. To me, this is clutter - maybe not today, but in a year or two, when there's tens of these quickstarts.

+0.

ge0ffrey:
It's a "factory layout" problem, which is in the supply chain sector, which is a sector with a lot of OptaPlanner use case and little OptaPlanner attention so far. It's also a good marketing effort.
Let's not fail this quickstart before we have given it a chance to live.

If in 1 to 2 years, we see that it doesn't pull its weight, we can remove it then. Quickstarts aren't API's: we don't commit to this remaining around if it doesn't have traction.

PR #20, opened 2020-11-02
gitattributes + minimize jbossNexus addition + use optaplanner-build-parent for multiproject

ge0ffrey:


triceo (approved):
No comments other than that the build is red.
But I've just merged a fix for that, so rerunning.

triceo:
Need to rerun again tomorrow, when the new SNAPSHOTs are built. Until then, this will be red.

ge0ffrey:
Rerunning

PR #28, opened 2020-11-23
Add content list and requirements

dupliaka:
Added list of content;
Since quick starts are mostly for developers and operators I suggest to concentrate on the technology rather then on the business problem itself.
Also I feel that it is quite important to underline the requirements needed to launch this demos.

### DIFF
--- a/README.adoc
+++ b/README.adoc
@@ -1,5 +1,21 @@
 = OptaPlanner Quickstarts
 
+This repository contains a growing collection of OptaPlanner quickstarts. 
+
+* Quarkus
+** <<School timetabling>> for Quarkus
+** <<Facility location problem>>
+** <<Factorio layout>>
+* SpringBoot 
+** <<School timetabling>> for SpringBoot
+* Kotlin
+** <<School timetabling>> for Kotlin
+
+== Java + Optaplanner Version Used
+
+Development branch will use the latest version of Java and Optaplanner.
+For now the quickstarts are tested with Java 15 and Optaplanner 8.
+   
 == Get started
 
 To get started with https://www.optaplanner.org/[OptaPlanner],


README.adoc
```diff
@@ -1,5 +1,21 @@
...
+This repository contains a growing collection of OptaPlanner quickstarts. 
+
+* Quarkus
+** <<School timetabling>> for Quarkus
+** <<Facility location problem>>
+** <<Factorio layout>>
+* SpringBoot 
+** <<School timetabling>> for SpringBoot
+* Kotlin
+** <<School timetabling>> for Kotlin
```
ge0ffrey, line 12: Isn't this the same information as what is below already? Or is this meant as a TOC (in which the links should work)

The "get started" should go first in, any case, I'd argue. Wdyt?

README.adoc
```diff
@@ -1,5 +1,21 @@
...
+
+== Java + Optaplanner Version Used
+
+Development branch will use the latest version of Java and Optaplanner.
+For now the quickstarts are tested with Java 15 and Optaplanner 8.
```
ge0ffrey, line 17: The quickstarts use java 11 by default. If we want to add such info, let's add it at the bottom.
First we need to convince them that OptaPlanner has use cases that resembles their needs and we need to get them started in less than 2 minutes. "real estate" at this part of the document is important.

README.adoc
ge0ffrey: Do we need a TOC of all quickstarts? Or does scrolling and seeing the screenshots work well enough?
Especially if we're going to have dozens of quickstarts over time.

README.adoc
ge0ffrey: Good point that we need to specify this JDK version requirement.
What if, at the end of the "getting started" section, we just add:
"Requires Java 11+ and Maven."?

README.adoc
dupliaka: Yes it is the same as all TOC does. I was thinking that it can help 
1) to group the quickstart in the meaning of technology since demos are mostly for developers
2) to have a full list of the growing amount of integration needed
Scrolling is great, but with all the demo images you cannot make a big picture of what we cover.

If the links are not working yet as supposed (just navigate to the right section in the same readme, please let me know)

dupliaka: Totally ok to have a get started first if it is a matter of time spend to try

README.adoc
dupliaka: Optaplanner maybe is not so well used with different versions to mention here, so that I will remove.

Ok, great idea, though it is a bit shame that we cannot write that it is tested with Java 15, because it shows that is new and cool(sorry for being naive). I wrote that this way because we got CI gob in git actions that are testing on Java 15. 
Maybe I got that wrong + we do not probably use all of the advantages of it now.

ge0ffrey:
@dupliaka Thank you for pointing out the issue of missing TOC and requirements. Thank for doing a PR. I do want to structure it a bit differently, so I 've taken your suggestions and incorporated them in this PR:
  https://github.com/kiegroup/optaplanner-quickstarts/pull/36

PR #29, opened 2020-11-24
PLANNER-2243 Release OptaPlanner 8.0.0.Final

rsynek:
We end up with 8.0.1-SNAPSHOT on the 8.0.x branch, this PR follows what the release automation will do.

### JIRA
https://issues.redhat.com/browse/PLANNER-2243

### DIFF
--- a/build/quickstarts-distribution/pom.xml
+++ b/build/quickstarts-distribution/pom.xml
@@ -5,7 +5,7 @@
   <parent>
     <groupId>org.optaplanner</groupId>
     <artifactId>optaplanner-quickstarts-parent</artifactId>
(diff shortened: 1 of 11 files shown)


ge0ffrey:
With the changes as a result from adding the quickstart showcase, distribution, I would except 3 other instances of 8.0.0-SNAPSHOT in the poms (usually the parent) to be replaced too?

rsynek:
Good catch, will need to look into it. The script your provided in https://issues.redhat.com/browse/PLANNER-2230 will need some tweaking before we automate this.

ge0ffrey (approved):

PR #30, opened 2020-11-27
Fix bat script + upgrade webjars

ge0ffrey:
UI's still look good after upgrade of webjars

### DIFF
--- a/build/quickstarts-distribution/src/main/assembly/scripts/runQuickstarts.bat
+++ b/build/quickstarts-distribution/src/main/assembly/scripts/runQuickstarts.bat
@@ -1,8 +1,8 @@
 @ECHO OFF
 setLocal enableExtensions enableDelayedExpansion
 
-jvmOptions="-Xms128m -Xmx512m -Dstartup-open-browser=true"
-mainJar="binaries/optaplanner-quickstarts-showcase-*-runner.jar"
+set "jvmOptions=-Xms128m -Xmx512m -Dstartup-open-browser=true"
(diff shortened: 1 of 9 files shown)


ge0ffrey:
This is not a blocker for 8.0, because in 8.0 we don't distribute the distribution yet and the "git clone" root bat script works fine.

build/quickstarts-distribution/src/main/assembly/scripts/runQuickstarts.bat
```diff
@@ -1,8 +1,8 @@
 @ECHO OFF
 setLocal enableExtensions enableDelayedExpansion
 
-jvmOptions="-Xms128m -Xmx512m -Dstartup-open-browser=true"
-mainJar="binaries/optaplanner-quickstarts-showcase-*-runner.jar"
+set "jvmOptions=-Xms128m -Xmx512m -Dstartup-open-browser=true"
+set "mainJar=binaries/optaplanner-quickstarts-showcase-${project.version}-runner.jar"
```
yurloc, line 5: @ge0ffrey Have you tried running this? Still doesn't work for me.

ge0ffrey: Thank you for noticing. I did make these change son the window side and then it ran, but testing this is very impractical. Something else must be off. I am going to merge this already because it's a step forward, and submit a new PR with full fixes for the windows experience.

PR #45, opened 2021-01-25
Remove comments for quarkus version update + branch process details

dupliaka:
Quarkus time tabling has gradle and maven builders. Versions in both should by synced only by contributors. Removed the comment

Removed information on removing repo for stable since that is related only to ours internal branch process.


rsynek (approved):
LGTM, thanks!

PR #50, opened 2021-02-03
Vaccine distribution prototype

ge0ffrey:
First implementation, as used in the video ( https://youtu.be/LTkoaBk-P6U )
I'll process the peer review comments tomorrow. Once it's all good, we 'll merge to the development branch
Given the urgency, I 'll also create a PR to the stable branch afterwards that cherry-picks/adjusts it using the latest Final version of optaplanner instead of snapshots.


kotlin-quarkus-school-timetabling/src/main/resources/META-INF/resources/app.js
```diff
@@ -60,6 +60,9 @@ function refreshTimeTable() {
             .append($(`<button type="button" class="ml-2 mb-1 btn btn-light btn-sm p-1"/>`)
               .append($(`<small class="fas fa-trash"/>`)
               ).click(() => deleteTimeslot(timeslot)))));
+      $.each(timeTable.roomList, (index, room) => {
+        rowByRoom.append($("<td/>").prop("id", `timeslot${timeslot.id}room${room.id}`));
+      });
```
triceo, line 65: Can we leave updates to other quickstarts to another PR, please?
I'm not saying we shouldn't do this, but let's separate concerns. This PR is big enough as it is.

quarkus-facility-location/src/main/java/org/acme/facilitylocation/domain/Location.java
```diff
@@ -18,9 +18,16 @@
...
 import com.fasterxml.jackson.annotation.JsonFormat;
 
+import static java.lang.Math.ceil;
+import static java.lang.Math.sqrt;
+
 @JsonFormat(shape = JsonFormat.Shape.ARRAY)
 public class Location {
 
+    // Approximate Metric Equivalents for Degrees. At the equator for longitude and for latitude anywhere,
+    // the following approximations are valid: 1° = 111 km (or 60 nautical miles) 0.1° = 11.1 km.
```
triceo, line 28: Is there any link to describe this in more detail anywhere?

quarkus-vaccination-scheduling/README.adoc
```diff
@@ -0,0 +1,70 @@
...
+[source, shell]
+----
+$ mvn quarkus:dev
+----
+
+. Visit http://localhost:8080 in your browser.
+
+. Click on the *Solve* button.
+
+Then try _live coding_:
```
triceo, line 16: Really? As the second thing the user sees?
I'd just remove this entire section, it takes attention away from the main point.

quarkus-vaccination-scheduling/src/main/java/org/acme/vaccinationscheduler/bootstrap/DemoDataGenerator.java
```diff
@@ -0,0 +1,145 @@
...
+
+import io.quarkus.runtime.StartupEvent;
+
+@ApplicationScoped
+public class DemoDataGenerator {
+
+    public static final double MINIMUM_LATITUDE = 33.40;
+    public static final double MAXIMUM_LATITUDE = 34.10;
+    public static final double MINIMUM_LONGITUDE = -84.90;
+    public static final double MAXIMUM_LONGITUDE = -83.90;
```
triceo, line 53: What is so special about these numbers? Perhaps a descriptive comment would be nice.

quarkus-vaccination-scheduling/src/main/java/org/acme/vaccinationscheduler/bootstrap/DemoDataGenerator.java
```diff
@@ -0,0 +1,145 @@
...
+    public VaccineType pickVaccineType(Random random) {
+        double randomDouble = random.nextDouble();
+        if (randomDouble < 0.50) {
+            return VaccineType.PFIZER;
+        } else if (randomDouble < 0.90) {
+            return VaccineType.MODERNA;
+        } else {
+            return VaccineType.ASTRAZENECA;
+        }
+    }
```
triceo, line 147: For easier customization, these thresholds would IMO look better on the enum directly.

quarkus-vaccination-scheduling/src/main/java/org/acme/vaccinationscheduler/domain/Injection.java
```diff
@@ -0,0 +1,93 @@
...
+
+    public VaccinationCenter getVaccinationCenter() {
+        return vaccinationCenter;
+    }
+
+    public int getLineIndex() {
+        return lineIndex;
+    }
+
+    public LocalDateTime getDateTime() {
```
triceo, line 86: The choice of `LocalDateTime` seems strange, considering we're possibly designing a system for entire countries, which may span several timezones.

quarkus-vaccination-scheduling/src/main/java/org/acme/vaccinationscheduler/domain/Location.java
```diff
@@ -0,0 +1,50 @@
...
+
+    public Location(double latitude, double longitude) {
+        this.latitude = latitude;
+        this.longitude = longitude;
+    }
+
+    public long getDistanceTo(Location other) {
+        double latitudeDiff = other.latitude - this.latitude;
+        double longitudeDiff = other.longitude - this.longitude;
+        return (long) ceil(sqrt(latitudeDiff * latitudeDiff + longitudeDiff * longitudeDiff) * METERS_PER_DEGREE);
```
triceo, line 42: Seems like a performance killer. 
Although the alternative of creating a massive distance matrix is probably overkill right here.

quarkus-vaccination-scheduling/src/main/java/org/acme/vaccinationscheduler/domain/Person.java
```diff
@@ -0,0 +1,89 @@
...
+
+    @PlanningId
+    private Long id;
+
+    private String name;
+    private Location homeLocation;
+    private LocalDate birthdate;
+    private int age;
+
+    private boolean firstShotInjected;
```
triceo, line 33: So a person with the first shot, and the person without the first shot, would be a different `Person` instance, even though they would technically be the same human?

I understand that this makes the model much simpler, but also doubles the memory requirements.

quarkus-vaccination-scheduling/src/main/java/org/acme/vaccinationscheduler/solver/VaccinationScheduleConstraintProvider.java
```diff
@@ -0,0 +1,114 @@
...
+//    Constraint elderlyFirstPerVaccinationCenter(ConstraintFactory constraintFactory) {
+//        Predicate<Injection> firstShotPredicate = (injection) -> !injection.getPerson().isFirstShotInjected();
+//        return constraintFactory
+//                .from(Injection.class).filter(firstShotPredicate)
+//                .join(constraintFactory.from(Injection.class).filter(firstShotPredicate),
+//                        Joiners.equal(Injection::getVaccinationCenter),
+//                        Joiners.greaterThan(injection -> injection.getPerson().getAge()),
+//                        Joiners.lessThan(Injection::getDateTime))
+//                .penalize("Elderly first per vaccination center", HardMediumSoftLongScore.ONE_SOFT);
+//    }
```
triceo, line 125: Removed for reasons of performance?
Because I do think that this constraint is actually pretty important.

triceo (approved):
Some comments on the model.
Did not review the JS parts.
Also have to look at the UI, but the comments on that will use a different channel than Github PR reviews.

triceo:
Also, let's add some test coverage, please.

quarkus-facility-location/src/main/java/org/acme/facilitylocation/domain/Location.java
Christopher-Chianelli: https://stackoverflow.com/a/25238112 goes into detail (and explains how to calculate it more accurately using the Haversine formula).

quarkus-vaccination-scheduling/src/main/resources/META-INF/resources/app.js
```diff
@@ -0,0 +1,286 @@
...
+      if (injection.person != null) {
+        personIdToInjectionMap.set(injection.person.id, injection);
+      }
+    });
+
+    const tbody = $(`<tbody>`).appendTo(scheduleTable);
+    var previousParsedDateTime = null;
+    solution.timeslotDateTimeList.forEach((dateTime) => {
+      const row = $(`<tr>`).appendTo(tbody);
+      var parsedDateTime = moment(dateTime, "YYYY,M,D,H,m");
```
Christopher-Chianelli, line 48: We may want to use another date time library: https://momentjs.com/docs/#/-project-status/

Moments are mutable, and it caused a fair bit of trouble for me before wondering why my dates are crazy
(cause: date.add() modifies the moment, not returning a new one).

Luxon is basically Moment but immutable (and is recommended by the moment team): https://moment.github.io/luxon/docs/manual/tour.html

quarkus-vaccination-scheduling/src/main/resources/application.properties
```diff
@@ -0,0 +1,14 @@
+########################
+# OptaPlanner properties
+########################
+
+# The solver runs for 30 seconds. To run for 30 seconds use "30s" and for 2 hours use "2h".
```
Christopher-Chianelli: Comment out of date; should be 5 minutes

Christopher-Chianelli (approved):
Minor comments; LGTM (reviewed both frontend and backend)

quarkus-facility-location/src/main/java/org/acme/facilitylocation/domain/Location.java
ge0ffrey: We might need to use Haversine, we use that in the TSP/VRP examples too, but it causes a performance overhead.
So far, the datasets run on the city of Atlanta are fine, but this needs to scale to the country level. Especially for countries like Canada and Sweden, and other near the poles, Haversine is going to make a difference.
Using real road traveling times is typically out of scope (think OSRM etc), given the urgency of deploying these kinds of use cases.

quarkus-vaccination-scheduling/src/main/java/org/acme/vaccinationscheduler/bootstrap/DemoDataGenerator.java
ge0ffrey: Those thresholds aren't an intrinsic part of the vaccine types. It's a simulation of the current reality, so I'd argue they belong to the demo generator?
In the real world cases, this won't matter, they'll replace the DemoGenerator

quarkus-vaccination-scheduling/src/main/java/org/acme/vaccinationscheduler/domain/Injection.java
ge0ffrey: As far as I can tell, no one has datasets that span multiple timezones. It's a business nightmare.
For example in the US, each state schedules their vaccines independently. It's even likely they split up states into regions for business reasons (which reduce our scaling reqs too, although we don't mind if they do it on a state level).

The quickstart should be simple. Countries that insist on doing scheduling across timezones can fork, switch to ZonedDateTime and enjoy worrying about DST. :)

quarkus-vaccination-scheduling/src/main/java/org/acme/vaccinationscheduler/domain/Location.java
ge0ffrey: You might be onto something. I'd call it a performance opportunity :) So far the score calculation count is 30k, well above 10k, is it doesn't kill it.
It all depends on the number of vaccination centers and the number of injectablePeople.
if `centerCount * injectablePeopleCount >= 100 000 000` then we need GB's of memory if we do a distance matrix.

Also the Haversine conversation comes into play here.

quarkus-vaccination-scheduling/src/main/java/org/acme/vaccinationscheduler/domain/Person.java
ge0ffrey: Currently the planning window isn't big enough to accommodate 2 shots for the same person. But if it grows to beyond 3 weeks, this is an issue indeed.

quarkus-vaccination-scheduling/src/main/java/org/acme/vaccinationscheduler/solver/VaccinationScheduleConstraintProvider.java
ge0ffrey: It doesn't kill performance, but it hurts it. (score calc from 30k to 5k). There's a better way to write it.

quarkus-vaccination-scheduling/src/main/resources/META-INF/resources/app.js
ge0ffrey: The mutability of momentjs has already bitten me. Great found Chris!
We have several other quickstarts using it too. Let's handle this luxon migration as a seperate issue after this is merged.

quarkus-vaccination-scheduling/README.adoc
ge0ffrey: The target audience is developers, live coding is one of the first things they need.
For the "I just want to demo this" audience, we have the runQuickstartsFromSource.sh script in the root of this repo and the instructions in the root readme (not this one) to run that.

quarkus-vaccination-scheduling/src/main/resources/META-INF/resources/app.js
ge0ffrey: Luxon is not yet avaiable on Maven Central as a simple webjar (it is as a npm webjar). Let's be a bit patient before using it. It does seem like the way to go forward in the future.

quarkus-facility-location/src/main/java/org/acme/facilitylocation/domain/Location.java
yurloc: It's a quick approximation. Taken from here https://www.usna.edu/Users/oceano/pguth/md_help/html/approx_equivalents.htm. It's basically the circumference divided by 360°. It's only accurate for latitudes close to equator.

quarkus-vaccination-scheduling/src/main/resources/META-INF/resources/app.js
yurloc: Built-in types from the [Intl](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Intl) namespace might be sufficient.

Also see https://momentjs.com/docs/#/-project-status/ why Moment.js is obsolete.

ge0ffrey:
> Also, let's add some test coverage, please.

Done :)

quarkus-vaccination-scheduling/src/main/java/org/acme/vaccinationscheduler/solver/VaccinationScheduleConstraintProvider.java
ge0ffrey: Added TODO, improvements will be for a next PR

quarkus-vaccination-scheduling/src/main/java/org/acme/vaccinationscheduler/domain/Person.java
ge0ffrey: The next PR or so will deal with continous planning and will take this issue into scope.

quarkus-vaccination-scheduling/src/main/java/org/acme/vaccinationscheduler/domain/Injection.java
ge0ffrey: Rejecting, see comment above.

quarkus-vaccination-scheduling/src/main/java/org/acme/vaccinationscheduler/bootstrap/DemoDataGenerator.java
ge0ffrey: Rejecting, see comment above.

quarkus-vaccination-scheduling/README.adoc
ge0ffrey: This discussion applies to all quickstarts - the existing ones do it like this too. Let's bring that discussion on zulip and pull it into a separate issue?

kotlin-quarkus-school-timetabling/src/main/resources/META-INF/resources/app.js
ge0ffrey: Yea, sorry about that. I 'll try to avoid it next time - it creates too much pain for the PR reviewers.

PR #56, opened 2021-02-09
Fix native build of showcase

rsynek:
The `surefire-plugin.version` was not defined for the showcase module. The `optaplanner-parent` defines `version.surefire.plugin` instead. This PR unifies the naming of plugin version properties to `version.{plugin name}.plugin`, which we already use in the `optaplanner` repository.

### DIFF
--- a/quarkus-facility-location/pom.xml
+++ b/quarkus-facility-location/pom.xml
@@ -8,10 +8,8 @@
   <version>1.0-SNAPSHOT</version>
 
   <properties>
-    <maven.compiler.parameters>true</maven.compiler.parameters>
     <maven.compiler.release>11</maven.compiler.release>
     <project.build.sourceEncoding>UTF-8</project.build.sourceEncoding>
-    <project.reporting.outputEncoding>UTF-8</project.reporting.outputEncoding>
 
     <version.io.quarkus>1.11.0.Final</version.io.quarkus>
     <version.org.optaplanner>8.3.0-SNAPSHOT</version.org.optaplanner>
--- a/quarkus-facility-location/pom.xml
+++ b/quarkus-facility-location/pom.xml
@@ -8,15 +8,16 @@
   <version>1.0-SNAPSHOT</version>
 
   <properties>
-    <version.org.optaplanner>8.3.0-SNAPSHOT</version.org.optaplanner>
-    <version.io.quarkus>1.11.0.Final</version.io.quarkus>
-
-    <compiler-plugin.version>3.8.1</compiler-plugin.version>
     <maven.compiler.parameters>true</maven.compiler.parameters>
     <maven.compiler.release>11</maven.compiler.release>
     <project.build.sourceEncoding>UTF-8</project.build.sourceEncoding>
     <project.reporting.outputEncoding>UTF-8</project.reporting.outputEncoding>
-    <surefire-plugin.version>2.22.1</surefire-plugin.version>
+
+    <version.io.quarkus>1.11.0.Final</version.io.quarkus>
+    <version.org.optaplanner>8.3.0-SNAPSHOT</version.org.optaplanner>
+
+    <version.compiler.plugin>3.8.1</version.compiler.plugin>
+    <version.surefire.plugin>2.22.1</version.surefire.plugin>
   </properties>
 
   <dependencyManagement>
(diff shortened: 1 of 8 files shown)


spring-boot-school-timetabling/pom.xml
```diff
@@ -8,14 +8,15 @@
...
 
   <properties>
+    <maven.compiler.target>11</maven.compiler.target>
+    <maven.compiler.source>11</maven.compiler.source>
+    <project.build.sourceEncoding>UTF-8</project.build.sourceEncoding>
+
     <version.org.optaplanner>8.3.0-SNAPSHOT</version.org.optaplanner>
     <version.org.springframework.boot>2.2.6.RELEASE</version.org.springframework.boot>
 
-    <project.build.sourceEncoding>UTF-8</project.build.sourceEncoding>
```
ge0ffrey, line 14: This is one is needed, the quickstarts don't extend optaplanner-parent (and they really shouldn't extend it either)

spring-boot-school-timetabling/pom.xml
```diff
@@ -8,14 +8,15 @@
   <version>1.0-SNAPSHOT</version>
 
   <properties>
+    <maven.compiler.target>11</maven.compiler.target>
+    <maven.compiler.source>11</maven.compiler.source>
```
ge0ffrey: Can we switch these two properties to one maven.compiler.release property too?

quarkus-facility-location/pom.xml
```diff
@@ -8,15 +8,16 @@
...
 
   <properties>
-    <version.org.optaplanner>8.3.0-SNAPSHOT</version.org.optaplanner>
-    <version.io.quarkus>1.11.0.Final</version.io.quarkus>
-
-    <compiler-plugin.version>3.8.1</compiler-plugin.version>
     <maven.compiler.parameters>true</maven.compiler.parameters>
     <maven.compiler.release>11</maven.compiler.release>
     <project.build.sourceEncoding>UTF-8</project.build.sourceEncoding>
     <project.reporting.outputEncoding>UTF-8</project.reporting.outputEncoding>
```
ge0ffrey: Let's remove project.reporting.outputEncoding, we're not using "mvn site".

quarkus-facility-location/pom.xml
```diff
@@ -8,15 +8,16 @@
   <version>1.0-SNAPSHOT</version>
 
   <properties>
-    <version.org.optaplanner>8.3.0-SNAPSHOT</version.org.optaplanner>
-    <version.io.quarkus>1.11.0.Final</version.io.quarkus>
-
-    <compiler-plugin.version>3.8.1</compiler-plugin.version>
     <maven.compiler.parameters>true</maven.compiler.parameters>
```
ge0ffrey: Check with Jiri if we really, really need this maven.compiler.parameters. If we do need this, add a comment why.
If we kinda need it, we don't really need it, so remove it, because for quickstarts the philosophy is "less is a lot more". Everything that is not use-case specific is guilty-until-proven-otherwise boilerplate code we'd love to remove.

ge0ffrey (approved):
Great work, see additional suggestions.

quarkus-facility-location/pom.xml
```diff
@@ -8,15 +8,14 @@
   <version>1.0-SNAPSHOT</version>
 
   <properties>
-    <version.org.optaplanner>8.3.0-SNAPSHOT</version.org.optaplanner>
-    <version.io.quarkus>1.11.0.Final</version.io.quarkus>
-
-    <compiler-plugin.version>3.8.1</compiler-plugin.version>
-    <maven.compiler.parameters>true</maven.compiler.parameters>
```
rsynek, line 15: @yurloc could you please confirm it's safe to delete this property? The quickstart seems to run just fine without it.

yurloc: It's generated by Quarkus Maven archetype. See https://quarkus.io/guides/hibernate-orm-panache for brief explanation. Since this quickstart doesn't use persistence, it's probably unused but I wouldn't remove it without verification.

ge0ffrey: none of the other quickstarts need it. Let's remove it until we prove that we need it :)

PR #57, opened 2021-02-09
Update snapshot of the release branch to the next micro version

dupliaka:
Pipeline automation was merged to master before 8.2.x
was created so it does not worked this time.


rsynek (changes requested):
Thanks for the upgrade, however, we need 2 commits and a tag:
1) upgrade to the .Final version
2) push the tag
3) force-push the `stable` branch to contains the `8.2.0.Final`
4) upgrade to the next micro snapshot

PR #59, opened 2021-02-10
Vaccination part 2: De-enum VaccinationType, add secondDoseReadyTime and IdealTime constraints, rename 2nd shot to 2nd dose

ge0ffrey:


triceo (approved):
LGTM.

Just FYI some of the comments from the previous round of review were not applied. Specifically the `LocalDate` usage, which I believe will become a problem if seriously used across the USA, where they have 3 different timezones, and sometimes even more than 1 inside 1 state.

ge0ffrey:
> Just FYI some of the comments from the previous round of review were not applied. Specifically the `LocalDate` usage, which I believe will become a problem if seriously used across the USA, where they have 3 different timezones, and sometimes even more than 1 inside 1 state.

We're not taking multiple timezones into scope for the quickstart. Non local date/time's make everything far more complex.

PR #65, opened 2021-02-25
[PLANNER-2349] Hide non-working hours and weekends in maintenance scheduling quickstart

cuijulian:
Jira: https://issues.redhat.com/browse/PLANNER-2349

Uses vis-timeline options to hide weekends and non-working hours outside of 9:00-17:00 in the maintenance schedule timeline.

### DIFF
--- a/quarkus-maintenance-scheduling/src/main/resources/META-INF/resources/app.js
+++ b/quarkus-maintenance-scheduling/src/main/resources/META-INF/resources/app.js
@@ -2,7 +2,7 @@ var autoRefreshCount = 0;
 var autoRefreshIntervalId = null;
 
 // Initial date/time (next Monday)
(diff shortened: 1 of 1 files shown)


triceo (approved):
LGTM.

One thing I just noticed: when I "zoom in" in the timeline by using the scroll wheel of my mouse, the view correctly zooms in. Unfortunately, it stays like that for a second or two, and then reverts to the original zoomed-out view without me doing anything.

It's not a blocker, but it is annoying.

cuijulian:
As discussed, the timeline auto-fits periodically during solving. There is a split-second delay between the engine stopping to solve and the UI showing it, so that's likely the "race-condition" issue you saw. Would it be better if the timeline didn't auto-zoom during solving? I don't feel strongly about either option. @ge0ffrey wdyt?

PR #68, opened 2021-03-03
PLANNER-2313 Create AMQ school timetabling quickstart

rsynek:
### JIRA
https://issues.redhat.com/browse/PLANNER-2313


amq-quarkus-school-timetabling/README.adoc
```diff
@@ -0,0 +1,139 @@
+= School Timetabling (Java, Quarkus, AMQ, Maven)
+
+The quickstart consists of two modules:
+
+* Client runs the UI and sends requests to the Solver via AMQ.
+* Solver optimizes the requests sent via AMQ and sends results back to AMQ broker.
+
+[NOTE]
+Starting the AMQ Artemis broker requires you to install https://docs.docker.com/compose/install/[docker-compose].
```
ge0ffrey: Instead of a note, can we make this a step in the next section?

Similar to how this section mentions step 3: Run a database (which the substep to download/install it)
https://github.com/kiegroup/optaplanner-quickstarts/blob/stable/quarkus-school-timetabling/README.adoc#run-a-native-executable

amq-quarkus-school-timetabling/README.adoc
```diff
@@ -0,0 +1,139 @@
...
+
+== Run the application with live coding
+
+. Start the AMQ Artemis broker and both Solver and Client application with a `run.sh` script:
```
ge0ffrey: Shall we refer to the "Solver" as the "Solver Server"? (for me "Solver" is Solver.java)

```suggestion
. Start the AMQ Artemis broker, the Solver Server and Client application with the `run.sh` script, which starts all 3 processes:
```

amq-quarkus-school-timetabling/README.adoc
```diff
@@ -0,0 +1,139 @@
...
+
+When you're done iterating in `quarkus:dev` mode, run the application as a conventional jar file.
+Before running any application, start the AMQ Artemis broker:
+
+[source, shell]
+----
+$ docker-compose up
+----
+
+=== Package and run the Solver application
```
ge0ffrey:
```suggestion
=== Package and run the Solver Server application
```

ge0ffrey: And so on below

amq-quarkus-school-timetabling/README.adoc
```diff
@@ -0,0 +1,139 @@
+= School Timetabling (Java, Quarkus, AMQ, Maven)
```
ge0ffrey: The directory is called "amq-quarkus", but this section header is called "..., Quarkus, AMQ, ..."

Following the Burr rule, the most important should be first. Here's that's AMQ I guess?
```suggestion
= School Timetabling (Java, AMQ, Quarkus, Maven)
```
That rule is a bit tricky though, as for Kotlin, the most important is Kotlin (not quarkus), but it works.

ge0ffrey: Motivation:
- A user looking for OptaPlanner on Quarkus should look at quarkus-school-timetabling
- A user looking for OptaPlanner on AMQ should look at amq-quarkus-school-timetabling
- A user looking for OptaPlanner on Kotlin should look at kotlin-quarkus-school-timetabling
- A user looking for OptaPlanner on Spring should look at spring-boot-school-timetabling

It works.

amq-quarkus-school-timetabling/amq-quarkus-school-timetabling-client/pom.xml
```diff
@@ -0,0 +1,118 @@
...
+<project xmlns="http://maven.apache.org/POM/4.0.0"
+         xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+         xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/xsd/maven-4.0.0.xsd">
+  <parent>
+    <artifactId>amq-quarkus-school-timetabling</artifactId>
+    <groupId>org.acme</groupId>
+    <version>1.0-SNAPSHOT</version>
+    <relativePath>../pom.xml</relativePath>
+  </parent>
+  <modelVersion>4.0.0</modelVersion>
```
ge0ffrey: Nitpick, we typically put modelVersion first, even before the parent element, see optaplanner-core

amq-quarkus-school-timetabling/amq-quarkus-school-timetabling-client/pom.xml
```diff
@@ -0,0 +1,118 @@
...
+      <groupId>io.quarkus</groupId>
+      <artifactId>quarkus-hibernate-orm-panache</artifactId>
+    </dependency>
+    <dependency>
+      <groupId>io.quarkus</groupId>
+      <artifactId>quarkus-jdbc-h2</artifactId>
+    </dependency>
+    <dependency>
+      <groupId>io.quarkus</groupId>
+      <artifactId>quarkus-hibernate-orm-rest-data-panache</artifactId>
```
ge0ffrey: Open question: Is the adding and deleting of lessons needed in scope?

In the beginning, I always added that in scope, but for other use cases (FLP, vaccination, maintenance scheduling), we deliberately started taking that out of scope to make the quickstarts slimmer (easier to consume). That being said, it's in there for all school-timetabling quickstarts, so it should be in here too.

amq-quarkus-school-timetabling/amq-quarkus-school-timetabling-client/pom.xml
```diff
@@ -0,0 +1,118 @@
...
+      <groupId>io.quarkus</groupId>
+      <artifactId>quarkus-resteasy</artifactId>
+    </dependency>
+    <dependency>
+      <groupId>io.quarkus</groupId>
+      <artifactId>quarkus-resteasy-jackson</artifactId>
+    </dependency>
+    <dependency>
+      <groupId>io.quarkus</groupId>
+      <artifactId>quarkus-hibernate-orm-panache</artifactId>
```
ge0ffrey: Open question: Does the client need database integration?
In the facility-location-problem, Jiri used no hiberante, no H2 and just keeps the problem in memory. The vaccination scheduling does the same.

I think we need (at least) one example with database integration for both quarkus, spring, and kotlin (quarkus-school-timetabling). I guess we also want to keep this example as much in sync with those variations? So let's keep it?

amq-quarkus-school-timetabling/amq-quarkus-school-timetabling-client/pom.xml
ge0ffrey: Same discussion as above. For consistency, let's keep it indeed.

amq-quarkus-school-timetabling/amq-quarkus-school-timetabling-client/src/main/java/org/acme/schooltimetabling/messaging/ErrorInfo.java
```diff
@@ -0,0 +1,39 @@
...
+ */
+
+package org.acme.schooltimetabling.messaging;
+
+public class ErrorInfo {
+    private String exceptionClassName;
+    private String exceptionMessage;
+
+    ErrorInfo() {
+        // Required for JSON deserialization.
```
ge0ffrey: Nitpick, in the other classes, we put the comment not in but before the constructor:
// No-arg constructor required for Jackson

amq-quarkus-school-timetabling/amq-quarkus-school-timetabling-client/src/main/java/org/acme/schooltimetabling/messaging/SolverJsonMapper.java
```diff
@@ -0,0 +1,33 @@
...
+import org.optaplanner.persistence.jackson.api.OptaPlannerJacksonModule;
+
+import com.fasterxml.jackson.databind.ObjectMapper;
+import com.fasterxml.jackson.datatype.jsr310.JavaTimeModule;
+
+public class SolverJsonMapper {
+
+    private static final ObjectMapper OBJECT_MAPPER = new ObjectMapper()
+            .registerModule(new JavaTimeModule())
+            .registerModule(OptaPlannerJacksonModule.createModule());
```
ge0ffrey: Important:
You shouldn't have to do this. Adding the quarkus-optaplanner-jackson extension should do this for you.

If that's for some reason not possible to reuse/inject that ObjectMapper, thats an issue we need to fix upstream.

amq-quarkus-school-timetabling/amq-quarkus-school-timetabling-client/src/main/java/org/acme/schooltimetabling/messaging/SolverRequest.java
```diff
@@ -0,0 +1,44 @@
...
+ * limitations under the License.
+ */
+
+package org.acme.schooltimetabling.messaging;
+
+import org.acme.schooltimetabling.domain.TimeTable;
+
+import io.quarkus.runtime.annotations.RegisterForReflection;
+
+@RegisterForReflection
```
ge0ffrey: Why is this needed?
Jackson does reflection, but it does so on Room.java too, which doesn't have this annotation? Maybe its a side affect that Room.java got registered by the solution hierachy registration (although Chris might be removing that with his solution cloner work). @Christopher-Chianelli wdyt?

amq-quarkus-school-timetabling/amq-quarkus-school-timetabling-client/src/main/java/org/acme/schooltimetabling/persistence/HardSoftScoreConverter.java
```diff
@@ -0,0 +1,44 @@
...
+
+package org.acme.schooltimetabling.persistence;
+
+import javax.persistence.AttributeConverter;
+import javax.persistence.Converter;
+
+import org.optaplanner.core.api.score.buildin.hardsoft.HardSoftScore;
+
+@Converter
+public class HardSoftScoreConverter implements AttributeConverter<HardSoftScore, String> {
```
ge0ffrey: Important: use optaplanner-persistence-jpa instead.

amq-quarkus-school-timetabling/amq-quarkus-school-timetabling-client/src/main/java/org/acme/schooltimetabling/rest/TimeTableResource.java
```diff
@@ -0,0 +1,97 @@
...
+    Emitter<String> solverRequestEmitter;
+
+    @Incoming("solver_response")
+    @Blocking
+    @Acknowledgment(Acknowledgment.Strategy.POST_PROCESSING)
+    public void process(String solverResponseMessage) {
+        SolverResponse solverResponse;
+        try {
+            solverResponse =
+                    SolverJsonMapper.get().readValue(solverResponseMessage, SolverResponse.class);
```
ge0ffrey: So normally, you should be able to inject the ObjectMapper I'd hope.

Actually, if your method signature is `process(SolverResponse ...)` I'd expect quarkus to automatically unmarshall that from json with that ObjectMapper, but you might need an annotation to tell Quarkus that the @Incoming is in json/application. something to ask on the quarkus zulip.

amq-quarkus-school-timetabling/amq-quarkus-school-timetabling-client/src/main/java/org/acme/schooltimetabling/messaging/SolverRequest.java
rsynek: Not only through the solution hierarchy; the REST extension registers for reflection any type you transfer via the REST API. This class is not part of the domain model nor does it participate in any REST endpoint. Thus, we have to use the annotation so that we can use the Jackson marshaling directly.

amq-quarkus-school-timetabling/amq-quarkus-school-timetabling-client/pom.xml
rsynek: +1 to the consistency: if we are ok with just the pre-generated dataset without any option to change it, let's remove them from every school-timetabling-based quickstart.

amq-quarkus-school-timetabling/amq-quarkus-school-timetabling-client/pom.xml
rsynek: Good question; similar to the REST endpoints, the motivation is to keep this quickstart consistent with the other school-timetabling variations. However, we pay for this consistency and so do our users: the quickstart is unnecessarily complex, although it should focus on its main topic.

amq-quarkus-school-timetabling/amq-quarkus-school-timetabling-client/src/main/java/org/acme/schooltimetabling/persistence/HardSoftScoreConverter.java
rsynek: Unfortunately, that's not possible at the moment with the GraalVM that comes in the builder image for native compilation. See https://github.com/kiegroup/optaplanner-quickstarts/pull/68/files#diff-3cf36f8484bac8df806e137ffca5bf14a6164e1144b2d3f32ff9d0eca8c6a97fR48-R50.

amq-quarkus-school-timetabling/amq-quarkus-school-timetabling-client/src/main/java/org/acme/schooltimetabling/messaging/SolverJsonMapper.java
rsynek: Good catch, injecting the `ObjectMapper` should be fine. Thanks.

amq-quarkus-school-timetabling/amq-quarkus-school-timetabling-client/src/main/java/org/acme/schooltimetabling/rest/TimeTableResource.java
rsynek: The automatic unmarshalling for the AMQP connector has its semantic described here:
https://smallrye.io/smallrye-reactive-messaging/smallrye-reactive-messaging/2.8/amqp/amqp.html#_deserialization

If we set the content type of the message to `application/json`, we get the `JsonObject` from `Vert.x`. In order to use the Jackson, we have to do the (un)marshalling ourselves.

amq-quarkus-school-timetabling/amq-quarkus-school-timetabling-client/src/main/java/org/acme/schooltimetabling/messaging/ErrorInfo.java
rsynek: Actually, I think this approach, although maybe inconsistent with other quickstarts, has two benefits:
- when you move the constructor, you always move it with the comment
- it makes it clearer that the constructor is intentionally empty

amq-quarkus-school-timetabling/amq-quarkus-school-timetabling-client/pom.xml
rsynek: I have been thinking about whether to keep the CRUD REST endpoints and the persistence. We try to maintain consistency wherever possible, but we also need to consider the value it brings.
The CRUD endpoints let users play with the quickstart for the cost of mere 3 marker interfaces => let's keep them.

amq-quarkus-school-timetabling/amq-quarkus-school-timetabling-client/pom.xml
rsynek: On second thoughts, persistence doesn't add anything beyond the basic functionality. However, it makes the quickstart more complex (extra code) and complicates the experience with the native compilation (property overriding, standalone H2 server). I don't think the value outweighs the cost and thus I am going to remove it.

amq-quarkus-school-timetabling/run.sh
```diff
@@ -0,0 +1,52 @@
...
+  echo "Waiting up to ${_timeout_seconds} seconds for ${_application_url} to become available."
+  while [[ "200" != $(curl -LI "${_application_url}" -o /dev/null -w '%{http_code}' -s) && ${_spent} -lt ${_timeout_seconds} ]]
+  do
+    sleep ${_increment}
+    _spent=$((_spent + _increment))
+  done
+}
+
+# Run docker-compose to start AMQ Artemis.
+docker-compose up > target/amq.log 2>&1 &
```
yurloc, line 38: Fail fast if `target/` doesn't exist.

amq-quarkus-school-timetabling/README.adoc
```diff
@@ -0,0 +1,138 @@
+= School Timetabling (Java, AMQ, Quarkus, Maven)
+
+The quickstart consists of two modules:
+
+* Client runs the UI and sends requests to the Solver Server via AMQ.
+* Solver Server optimizes the requests sent via AMQ and sends results back to AMQ broker.
+
+== Run the application with live coding
+
```
yurloc, line 9: Run `mvn clean install`.

amq-quarkus-school-timetabling/run.sh
```diff
@@ -0,0 +1,52 @@
...
+readonly PID_DOCKER_AMQ=$!
+echo "Running the AMQ broker via docker."
+
+# Run the solver.
+mvn clean quarkus:dev -f amq-quarkus-school-timetabling-solver -Dquarkus.http.port=$SOLVER_PORT -Ddebug=8180 > target/solver.log 2>&1 &
```
yurloc: SC2086: Double quote to prevent globbing and word splitting.

amq-quarkus-school-timetabling/amq-quarkus-school-timetabling-solver/src/main/java/org/acme/schooltimetabling/messaging/TimeTableMessagingHandler.java
```diff
@@ -0,0 +1,106 @@
...
+    @Inject
+    ObjectMapper objectMapper;
+
+    @Inject
+    @Channel(SOLVER_RESPONSE_CHANNEL)
+    Emitter<String> solverResponseEmitter;
+
+    @Inject
+    TimeTableMessagingHandler(SolverFactory<TimeTable> solverFactory) {
+        solver = solverFactory.buildSolver();
```
yurloc, line 59: Why can't you simply inject a solver instance?

amq-quarkus-school-timetabling/amq-quarkus-school-timetabling-solver/src/main/java/org/acme/schooltimetabling/domain/Lesson.java
```diff
@@ -0,0 +1,101 @@
...
+ */
+
+package org.acme.schooltimetabling.domain;
+
+import org.optaplanner.core.api.domain.entity.PlanningEntity;
+import org.optaplanner.core.api.domain.lookup.PlanningId;
+import org.optaplanner.core.api.domain.variable.PlanningVariable;
+
+@PlanningEntity
+public class Lesson {
```
yurloc: Why are there same classes (FQN-wise) with a lot of duplicate code? That's not an example of a best practice. In my opinion, a shared, domain module would make the example clearer and by demonstrating how to reuse code for data that is transferred between different services it would prevent people replicating the current approach with code duplication.

amq-quarkus-school-timetabling/run.sh
rsynek: good catch, thanks

amq-quarkus-school-timetabling/amq-quarkus-school-timetabling-solver/src/main/java/org/acme/schooltimetabling/messaging/TimeTableMessagingHandler.java
rsynek: The optaplanner-quarkus extension does not provide any `Solver` bean; it provides the `SolverManager`, `ScoreManager`, and `SolverFactory`. Given the CDI in Quarkus creates the beans lazily, it shouldn't be an issue to create one.

amq-quarkus-school-timetabling/amq-quarkus-school-timetabling-solver/src/main/java/org/acme/schooltimetabling/domain/Lesson.java
rsynek: The intention is to avoid mixing JPA and OptaPlanner in the same classes, as these two concerns belong to separate modules (Client, Solver).
Now, if I remove the persistence from the Client (see the other comments above), it opens the option to remove this duplication.

amq-quarkus-school-timetabling/amq-quarkus-school-timetabling-client/pom.xml
rsynek: Unfortunately, there is no easy to keep it without persistence - it's very tightly coupled.

amq-quarkus-school-timetabling/amq-quarkus-school-timetabling-client/src/main/java/org/acme/schooltimetabling/rest/TimeTableResource.java
```diff
@@ -0,0 +1,95 @@
...
+        if (solverResponse.isSuccess()) {
+            TimeTable timeTable = solverResponse.getTimeTable();
+            timeTable.setSolverStatus(SolverStatus.NOT_SOLVING);
+            timeTableRepository.update(timeTable);
+        } else {
+            timeTableRepository.get().setSolverStatus(SolverStatus.NOT_SOLVING);
+            throw new IllegalStateException("Error during solving. "
+                    + solverResponse.getErrorInfo().getExceptionClassName()
+                    + " : "
+                    + solverResponse.getErrorInfo().getExceptionMessage());
```
ge0ffrey: Should we apply our error messages style so it's clear what is the exception class and the message?
```
   throw new IllegalStateException("Solving failed with exception class ("
                    + solverResponse.getErrorInfo().getExceptionClassName()
                    + ") and message (" + solverResponse.getErrorInfo().getExceptionMessage() + ").");
```

amq-quarkus-school-timetabling/amq-quarkus-school-timetabling-client/src/main/resources/application.properties
```diff
@@ -0,0 +1,40 @@
...
+# timeTable.demoData=LARGE
+
+########################
+# Messaging
+########################
+
+amqp-username=quarkus
+amqp-password=quarkus
+
+# Configure the Kafka sink (we write to it)
```
ge0ffrey: Avoid "we":
```suggestion
# Configure the Kafka sink to write to
```

amq-quarkus-school-timetabling/amq-quarkus-school-timetabling-client/src/main/resources/application.properties
```diff
@@ -0,0 +1,40 @@
...
+mp.messaging.outgoing.solver_request.connector=smallrye-amqp
+mp.messaging.outgoing.solver_request.durable=true
+
+# Configure the Kafka source (we read from it)
+mp.messaging.incoming.solver_response.connector=smallrye-amqp
+mp.messaging.incoming.solver_response.durable=true
+
+# TODO: Remove after switching to fast jar by default.
```
ge0ffrey: This will be obsolete in Quarkus 1.12+. Can the TODO "expiry condition" be clearer by saying "Remove after upgrading to Quarkus 1.12 or higher which uses fast-jar by default"?

That way, it sticks out more like a soar thumb when we are on Quarkus 1.14 in a 6 months from now and someone happens to look at this code.

amq-quarkus-school-timetabling/amq-quarkus-school-timetabling-client/src/test/java/org/acme/schooltimetabling/TimeTableResourceTest.java
```diff
@@ -0,0 +1,121 @@
...
+
+    @BeforeEach
+    void setupChannels() {
+        solverRequestSink = connector.sink("solver_request");
+        solverResponseSource = connector.source("solver_response");
+        prepareTestTimeTable();
+    }
+
+    @Test
+    @Timeout(60L)
```
ge0ffrey, line 84: 60 seconds can be short on some CI machines sometimes.

For the timeouts on optaplanner-example's integration tests (such as CloudBalancingPerformanceTest), we set 600 seconds, because it runs within 60 seconds (actually on average 10 seconds) on our machines.

Therefore, to avoid false negatives, I recommending taking the max you think it can be and then multiply that by 10. The timeout only matters when that particular test fails, so that will be very rare anyway :)

ge0ffrey: Or put differently: 1 false negative is worse for our team than 10 times taking 9 extra minutes to fail, because false negatives make us mistrust our CI.

amq-quarkus-school-timetabling/amq-quarkus-school-timetabling-solver/src/main/java/org/acme/schooltimetabling/messaging/TimeTableMessagingHandler.java
```diff
@@ -0,0 +1,109 @@
...
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+import com.fasterxml.jackson.core.JsonProcessingException;
+import com.fasterxml.jackson.databind.ObjectMapper;
+
+@ApplicationScoped
+public class TimeTableMessagingHandler {
+
+    private static final Logger LOGGER = LoggerFactory.getLogger(TimeTableMessagingHandler.class);
```
ge0ffrey, line 44: Can we standarize this naming across the quickstarts? (=> separate issue probably)
QuickstartLauncherResource and optaplanner-examples typically call it `logger`. But WikiConnectorResource calls it `LOG`.

amq-quarkus-school-timetabling/amq-quarkus-school-timetabling-solver/src/main/java/org/acme/schooltimetabling/messaging/TimeTableMessagingHandler.java
```diff
@@ -0,0 +1,109 @@
...
+
+    @Inject
+    TimeTableMessagingHandler(SolverFactory<TimeTable> solverFactory) {
+        solver = solverFactory.buildSolver();
+    }
+
+    @Incoming(SOLVER_REQUEST_CHANNEL)
+    public CompletionStage<Void> solve(Message<String> solverRequestMessage) {
+        return CompletableFuture.runAsync(() -> {
+            SolverRequest solverRequest = null;
```
ge0ffrey: (nitpick) I don't `= null` if it isn't needed (the code either fills in correctly or the catch returns or throws.
```suggestion
            SolverRequest solverRequest;
```

amq-quarkus-school-timetabling/amq-quarkus-school-timetabling-solver/src/main/java/org/acme/schooltimetabling/messaging/TimeTableMessagingHandler.java
```diff
@@ -0,0 +1,109 @@
...
+            try {
+                solverRequest = objectMapper.readValue(solverRequestMessage.getPayload(), SolverRequest.class);
+            } catch (JsonProcessingException ex) {
+                LOGGER.warn("Unable to deserialize solver request from JSON.", ex);
+                // Bad request should go directly to a DLQ.
+                solverRequestMessage.nack(ex);
+                return;
```
ge0ffrey, line 74: (important)
Normally, when things go wrong, we fail fast: we throw an exception instead of doing "warn" and returning, as this can lead to the exception being ignored. Especially as JsonProcessingException is - unlike a "wrong password" exception - something that shouldn't happen normally.

If there's a good reason why we can't just throw an exception (it brings the whole system down - and that would be a bad thing - usually that is a good thing so the bug get fixed), so the warn makes sense, let's add a comment why this case is exceptional.

ge0ffrey: Ok, reading further it makes 100% sense why we don't bring down the whole system here, as the client makes a mistake. But I would except that if you do throw an exception here, that quarkus won't bring down the whole system.
Just like the swing event thread and http servlet thread have a "don't-kill-the-thread" general exception handler, I 'd expect the quarkus reactive thing to have this too. Something to try out.

If that's there, we can nack the message and throw the exception, without the log line.
Why is that better? Because any line could throw an exception.  `objectMapper.readValue` could throw an exception that is not caught by JsonProcessingException. solver.solve could throw an Error that is not caught by Exception.

amq-quarkus-school-timetabling/amq-quarkus-school-timetabling-solver/src/main/java/org/acme/schooltimetabling/messaging/TimeTableMessagingHandler.java
```diff
@@ -0,0 +1,109 @@
...
+            }
+
+            TimeTable solution = null;
+            try {
+                solution = solver.solve(solverRequest.getTimeTable());
+                replySuccess(solverRequestMessage, solverRequest.getProblemId(), solution);
+            } catch (Exception ex) {
```
ge0ffrey: Don't catch Exception, that's an anti-pattern, unless it's a "don't-kill-the-thread" exception handler, in which case it should probably be catching Throwable. See above

ge0ffrey: Hmm, whatever happens, we do want to send a signal to the client. Even if it's an OutOfMemoryError in the json parsing.  So it sounds like we need an exception handler. Let's discuss.

amq-quarkus-school-timetabling/amq-quarkus-school-timetabling-client/src/test/java/org/acme/schooltimetabling/TimeTableResourceTest.java
rsynek: That's exactly the story behind the test timeout here: the 60 seconds is already generous, as the test usually finishes in ~1 second.

amq-quarkus-school-timetabling/amq-quarkus-school-timetabling-client/src/main/java/org/acme/schooltimetabling/persistence/HardSoftScoreConverter.java
rsynek: Irrelevant due to dropping the persistence.

ge0ffrey (approved):

rsynek:
Please wait with merging, I want to address the other comments first.

amq-quarkus-school-timetabling/amq-quarkus-school-timetabling-solver/src/main/java/org/acme/schooltimetabling/messaging/TimeTableMessagingHandler.java
rsynek: Agreed to try to catch even the throwable (no guarantees about the success in case e.g. an OOME).

amq-quarkus-school-timetabling/amq-quarkus-school-timetabling-solver/src/main/java/org/acme/schooltimetabling/messaging/TimeTableMessagingHandler.java
rsynek: If we take a look also at other modules of the optaplanner repository, we get `LOGGER` and `log` as well.

Here is the issue: https://issues.redhat.com/browse/PLANNER-2368

amq-quarkus-school-timetabling/amq-quarkus-school-timetabling-solver/src/main/java/org/acme/schooltimetabling/messaging/TimeTableMessagingHandler.java
rsynek: The exception handling you're suggesting works only with automatic acknowledgment. Here we use manual acknowledgment via the `Message` class and thus we are responsible for calling either `ack()` or `nack()`.
I am improving the comment explaining this piece of code.

amq-quarkus-school-timetabling/amq-quarkus-school-timetabling-solver/src/main/java/org/acme/schooltimetabling/messaging/TimeTableMessagingHandler.java
rsynek: https://issues.redhat.com/browse/PLANNER-2369

PR #72, opened 2021-03-10
PLANNER-2345: Autowire ConstraintVerifier in Spring Boot tests

Christopher-Chianelli:
### JIRA

https://issues.redhat.com/browse/PLANNER-2345


triceo (approved):

triceo:
@Christopher-Chianelli Will merge when the tests are green.

Christopher-Chianelli:
@triceo does the hooks use the latest optaplanner master (or only the latest snapshot)? It seem it doesn't have the latest changes yet (works on my machine). Probably need to rerun CI tomorrow.

triceo:
If it's the same hook as with core, it is using SNAPSHOTs from Maven. So yes, it will need to be rerun tomorrow.
However, if we had the internal Jenkins here, that would already have the right SNAPSHOT.

PR #81, opened 2021-03-18
BXMSPROD-1255 CI flows

Ginxo:
### JIRA
https://issues.redhat.com/browse/BXMSPROD-1255

### Referenced pull requests

- https://github.com/kiegroup/optaplanner-quickstarts/pull/81
- https://github.com/kiegroup/optaplanner-quickstarts/pull/82
- https://github.com/kiegroup/optaplanner-quickstarts/pull/86

### DIFF
--- a/.ci/project-dependencies.yaml
+++ b/.ci/project-dependencies.yaml
@@ -4,4 +4,11 @@ dependencies:
 
   - project: kiegroup/optaplanner-quickstarts
     dependencies:
-      - project: kiegroup/optaplanner
+      - project: kiegroup/optaplanner    
+    mapping:
+      dependencies:
+        default:
+          source: development
+          target: master
+      source: master
+      target: development
--- a/.ci/project-dependencies.yaml
+++ b/.ci/project-dependencies.yaml
@@ -1,0 +1,14 @@
+version: "2.0"
+dependencies:
+  - project: kiegroup/optaplanner
+
+  - project: kiegroup/optaplanner-quickstarts
+    dependencies:
+      - project: kiegroup/optaplanner
+    mapping:
+      dependencies:
+        default:
+          source: development
+          target: master
+      source: master
+      target: development
--- a/.ci/project-dependencies.yaml
+++ b/.ci/project-dependencies.yaml
@@ -5,10 +5,3 @@ dependencies:
   - project: kiegroup/optaplanner-quickstarts
     dependencies:
       - project: kiegroup/optaplanner
-    mapping:
-      dependencies:
-        default:
-          source: development
-          target: master
-      source: master
-      target: development
--- a/.ci/pull-request-config.yaml
+++ b/.ci/pull-request-config.yaml
@@ -5,8 +5,8 @@ dependencies: ./project-dependencies.yaml
 build:
   - project: kiegroup/optaplanner
     build-command:
-      upstream: mvn -e clean install -DskipTests -Dgwt.compiler.skip=true -Dgwt.skipCompilation=true -Denforcer.skip=true -Dcheckstyle.skip=true -Dspotbugs.skip=true -Drevapi.skip=true -Dmaven.wagon.httpconnectionManager.ttlSeconds=25 -Dmaven.wagon.http.retryHandler.count=3
+      upstream: mvn -e clean install -Dquickly -Dmaven.wagon.httpconnectionManager.ttlSeconds=25 -Dmaven.wagon.http.retryHandler.count=3
 
   - project: kiegroup/optaplanner-quickstarts
     build-command:
-      current: mvn -e -nsu -Dfull -Pwildfly clean install -Prun-code-coverage  -Dcontainer.profile=wildfly -Dcontainer=wildfly -Dintegration-tests=true -Dmaven.test.failure.ignore=true -Dmaven.wagon.httpconnectionManager.ttlSeconds=25 -Dmaven.wagon.http.retryHandler.count=3
+      current: mvn -e -nsu -Dfull clean install --fail-at-end -Dmaven.wagon.httpconnectionManager.ttlSeconds=25 -Dmaven.wagon.http.retryHandler.count=3
--- a/.ci/pull-request-config.yaml
+++ b/.ci/pull-request-config.yaml
@@ -1,0 +1,12 @@
+version: "2.0"
+
+dependencies: ./project-dependencies.yaml
+
+build:
+  - project: kiegroup/optaplanner
+    build-command:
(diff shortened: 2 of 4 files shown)


mareknovotny:
@ge0ffrey @triceo @rsynek your review is required as this branch was not setup for any CI

mareknovotny (approved):

antncl (approved):

.ci/pull-request-config.yaml
```diff
@@ -0,0 +1,12 @@
+version: "2.0"
+
+dependencies: ./project-dependencies.yaml
+
+build:
+  - project: kiegroup/optaplanner
+    build-command:
+      upstream: mvn -e clean install -DskipTests -Dgwt.compiler.skip=true -Dgwt.skipCompilation=true -Denforcer.skip=true -Dcheckstyle.skip=true -Dspotbugs.skip=true -Drevapi.skip=true -Dmaven.wagon.httpconnectionManager.ttlSeconds=25 -Dmaven.wagon.http.retryHandler.count=3
```
rsynek: Please remove any GWT-related properties; this technology is no longer used by any optaplanner/optaweb project.

As regard to skipping the other plugins, maybe you want to use the [quick profile](https://github.com/kiegroup/optaplanner/blob/master/build/optaplanner-build-parent/pom.xml#L1128) ?

.ci/project-dependencies.yaml
```diff
@@ -0,0 +1,14 @@
...
+dependencies:
+  - project: kiegroup/optaplanner
+
+  - project: kiegroup/optaplanner-quickstarts
+    dependencies:
+      - project: kiegroup/optaplanner
+    mapping:
+      dependencies:
+        default:
+          source: development
```
rsynek, line 11: Just to be sure: is it ok to put here the `development` branch, as this PR targets the branch `8.3.x`?

rsynek (commented):
LGTM, a couple of suggestions to simplify the maven command.

.ci/pull-request-config.yaml
mareknovotny: but we do want to run Integration tests or not for this branch?

rsynek: This is the upstream repo's maven build command, for which, as I understand, we skip tests. Am I missing something?

mbiarnes (approved):
LGTM

.ci/pull-request-config.yaml
mareknovotny: @rsynek true it is for upstream, sorry to not catch it in the context.

.ci/project-dependencies.yaml
Ginxo: @rsynek then no mapping required here, right?

.ci/pull-request-config.yaml
Ginxo: @rsynek done!

.ci/project-dependencies.yaml
rsynek: The branching scheme is as follows:
optaplanner-quickstarts -> optaplanner
`development`               -> `master`
`8.5.x`                            -> `8.5.x`
`8.4.x`                            -> `8.4.x`
...

mareknovotny (approved):

.ci/pull-request-config.yaml
```diff
@@ -0,0 +1,12 @@
+version: "2.0"
+
+dependencies: ./project-dependencies.yaml
+
+build:
+  - project: kiegroup/optaplanner
+    build-command:
+      upstream: mvn -e clean install -DskipTests -Denforcer.skip=true -Dcheckstyle.skip=true -Dspotbugs.skip=true -Drevapi.skip=true -Dmaven.wagon.httpconnectionManager.ttlSeconds=25 -Dmaven.wagon.http.retryHandler.count=3
```
rsynek: I checked the skip properties we pass here with the `quick` profile introduced exactly for this purpose and suggest replacing them with `-Dquickly`, which activates the profile. The `maven.wagon.*` properties have been removed as they are already present in https://github.com/kiegroup/optaplanner/blob/master/.mvn/maven.config. That being said, if you think the values here work better, please do override the maven.config file. I would like to keep this tweaking in a single place.

```suggestion
      upstream: mvn -e clean install -Dquickly
```

.ci/pull-request-config.yaml
```diff
@@ -0,0 +1,12 @@
...
+
+  - project: kiegroup/optaplanner-quickstarts
+    build-command:
+      current: mvn -e -nsu -Dfull -Pwildfly clean install -Prun-code-coverage  -Dcontainer.profile=wildfly -Dcontainer=wildfly -Dintegration-tests=true -Dmaven.test.failure.ignore=true -Dmaven.wagon.httpconnectionManager.ttlSeconds=25 -Dmaven.wagon.http.retryHandler.count=3
```
rsynek: Removing irrelevant profiles and properties.
```suggestion
      current: mvn -e -nsu -Dfull clean install -Dmaven.test.failure.ignore=true -Dmaven.wagon.httpconnectionManager.ttlSeconds=25 -Dmaven.wagon.http.retryHandler.count=3
```

rsynek: Why do we need the `maven.test.failure.ignore=true` ?

mareknovotny: i think it was to finish all modules if possible to show ALL failures. In case of an error in build it will stop immediately while on test failures it makes sense to continue with following tests and modules until end.

rsynek: I think that what you are describing is `--fail-at-end`. Maybe the property `maven.test.failure.ignore=true` never fails and it's up to Jenkins to interpret the test results?
Unless there is a good reason to use the property, I would rather go with the `--fail-at-end` option.

.ci/pull-request-config.yaml
Ginxo: @rsynek thanks, I would keep `-Dmaven.wagon.httpconnectionManager.ttlSeconds=25 -Dmaven.wagon.http.retryHandler.count=3` to avoid problem with GA+maven.
Let me merge your proposal with mine...

Ginxo: done

.ci/pull-request-config.yaml
Ginxo: changed to `--fail-at-end`

Ginxo:
@rsynek changes applied, please review and once we have this ok I will apply same approach for `8.4.x` `8.5.x` and `development` (development will require mapping development -> master)

rsynek (approved):

Ginxo:
- [UPDATED] https://github.com/kiegroup/optaplanner-quickstarts/pull/81 (mapping applied to keep consistency) 
- [UPDATED] https://github.com/kiegroup/optaplanner-quickstarts/pull/82 (changes from #81 applied)
- [CREATED] https://github.com/kiegroup/optaplanner-quickstarts/pull/86


please @rsynek @mbiarnes @mareknovotny review/merge thanks

PR #85, opened 2021-03-23
[PLANNER-2350] Show ready + due date for maintenance jobs in separate manager timeline + safety margin

cuijulian:
Jira: https://issues.redhat.com/browse/PLANNER-2350

This PR adds a manager timeline view showing ready/due dates and safety margins for each maintenance job.
WIP until the feasible solution test error is fixed.

### DIFF
--- a/quarkus-maintenance-scheduling/src/main/java/org/acme/maintenancescheduling/domain/MaintenanceJob.java
+++ b/quarkus-maintenance-scheduling/src/main/java/org/acme/maintenancescheduling/domain/MaintenanceJob.java
@@ -16,20 +16,14 @@
 
 package org.acme.maintenancescheduling.domain;
 
-import org.optaplanner.core.api.domain.entity.PlanningEntity;
-import org.optaplanner.core.api.domain.lookup.PlanningId;
-import org.optaplanner.core.api.domain.variable.PlanningVariable;
-
 import javax.persistence.Entity;
 import javax.persistence.GeneratedValue;
 import javax.persistence.Id;
 import javax.persistence.ManyToOne;
 
 @Entity
-@PlanningEntity
 public class MaintenanceJob {
 
-    @PlanningId
     @Id
     @GeneratedValue
     private Long id;
@@ -38,19 +32,10 @@ public class MaintenanceJob {
 
     @ManyToOne
     private MaintainableUnit maintainableUnit;
-    
+
     private int readyTimeGrainIndex;
     private int dueTimeGrainIndex;
     private int durationInGrains;
-
-    // TODO: Add configuration option for how long each TimeGrain is
-    @PlanningVariable(valueRangeProviderRefs = "timeGrainRange")
-    @ManyToOne
-    private TimeGrain startingTimeGrain;
-
-    @PlanningVariable(valueRangeProviderRefs = "assignedCrewRange")
-    @ManyToOne
-    private MaintenanceCrew assignedCrew;
 
     private int safetyMarginDurationInGrains;
 
@@ -61,8 +46,8 @@ public class MaintenanceJob {
     }
 
     public MaintenanceJob(String jobName, MaintainableUnit maintainableUnit, int readyTimeGrainIndex,
-                          int dueTimeGrainIndex, int durationInGrains, int safetyMarginDurationInGrains, 
-                          boolean critical) {
+                                    int dueTimeGrainIndex, int durationInGrains, int safetyMarginDurationInGrains,
+                                    boolean critical) {
         this.jobName = jobName;
         this.maintainableUnit = maintainableUnit;
         this.readyTimeGrainIndex = readyTimeGrainIndex;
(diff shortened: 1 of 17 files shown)


quarkus-maintenance-scheduling/src/main/java/org/acme/maintenancescheduling/solver/MaintenanceScheduleConstraintProvider.java
```diff
@@ -125,5 +126,14 @@ public Constraint assignAllNonCriticalJobs(ConstraintFactory constraintFactory)
     // TODO: Risk: when job completion date falls within the “SafetyMargin” before the due date,
     //  square the time between the start of the “SafetyMargin” and completion date, sum for all jobs
     //  (quadratic, will scale faster. Sq root the sum of squares)
+    public Constraint jobsShouldFinishBeforeSafetyMargin(ConstraintFactory constraintFactory) {
+        return constraintFactory.fromUnfiltered(MaintenanceJob.class)
+                .filter(maintenanceJob -> maintenanceJob.getStartingTimeGrain() != null
+                        && maintenanceJob.getStartingTimeGrain().getGrainIndex()
+                                + maintenanceJob.getDurationInGrains() > (maintenanceJob.getDueTimeGrainIndex() - maintenanceJob.getSafetyMarginDurationInGrains()))
```
triceo: This logic deserves to be extracted to its own method.
First of all, it will help with understanding the constraint at first look.
Second, you're using the same code twice.

quarkus-maintenance-scheduling/src/main/java/org/acme/maintenancescheduling/domain/MaintenanceJob.java
```diff
@@ -172,6 +174,14 @@ public void setAssignedCrew(MaintenanceCrew assignedCrew) {
         this.assignedCrew = assignedCrew;
     }
 
+    public int getSafetyMarginDurationInGrains() {
+        return safetyMarginDurationInGrains;
+    }
+
+    public void setSafetyMarginDurationInGrains(int safetyMarginDurationInGrains) {
+        this.safetyMarginDurationInGrains = safetyMarginDurationInGrains;
+    }
```
triceo: How did you decide to calculate the safety margin?

quarkus-maintenance-scheduling/src/main/java/org/acme/maintenancescheduling/solver/MaintenanceScheduleConstraintProvider.java
```diff
@@ -125,5 +126,14 @@ public Constraint assignAllNonCriticalJobs(ConstraintFactory constraintFactory)
...
+                .penalizeConfigurable("Jobs should finish before safety margin",
+                        maintenanceJob -> maintenanceJob.getStartingTimeGrain().getGrainIndex()
+                                + maintenanceJob.getDurationInGrains() - (maintenanceJob.getDueTimeGrainIndex() - maintenanceJob.getSafetyMarginDurationInGrains()));
+    }
```
triceo: Should we perhaps weight this more heavily towards the end of the margin?
(Not sure, just starting a discussion.)

triceo (commented):
First round of review finished.
I will likely have some more comments at/after the meeting today.

quarkus-maintenance-scheduling/src/main/java/org/acme/maintenancescheduling/domain/MutuallyExclusiveJobs.java
```diff
@@ -25,6 +27,8 @@
 import java.util.List;
 
 @Entity
+// TODO: Temporary fix to deep clone problem fact since it references a planning entity (MaintenanceJob)
```
triceo: The actual fix is to make sure that the `Job` is not an entity. Instead, something like `MaintenanceJobAssignment` becomes the entity, and it points to the `Job` which becomes a pure fact.

triceo: We should prioritize the fix, because this data model is wrong and we should not be exposing users to it anymore.
I am sorry I did not notice this.

cuijulian: That makes sense. Just so I understand, what is the antipattern in the current model? Is it the fact that a planning entity (`Job`) is a field in a problem fact (`MutuallyExclusiveJobs`)? How do we tell users this is an antipattern? Fail fast seems to be the current consensus in the Zulip discussion.

Interestingly, the OptaPlanner docs mentions the workaround implemented here: `If any of your problem facts needs to be deep cloned for a planning clone, for example if the problem fact references a planning entity or the planning solution, mark its class with a @DeepPlanningClone annotation`: https://docs.optaplanner.org/8.4.1.Final/optaplanner-docs/html_single/index.html#fieldAccessingSolutionCloner

triceo: > That makes sense. Just so I understand, what is the antipattern in the current model? Is it the fact that a planning entity (`Job`) is a field in a problem fact (`MutuallyExclusiveJobs`)?

Correct.

> Interestingly, the OptaPlanner docs mentions the workaround implemented here: `If any of your problem facts needs to be deep cloned for a planning clone, for example if the problem fact references a planning entity or the planning solution, mark its class with a @DeepPlanningClone annotation`: https://docs.optaplanner.org/8.4.1.Final/optaplanner-docs/html_single/index.html#fieldAccessingSolutionCloner

Yes. That said, deep cloning is expensive, and if we can avoid it, good for us.

PR #96, opened 2021-04-15
PLANNER-2362 Realtime planning quickstart: domain model, constraints, UI

rsynek:
### JIRA
https://issues.redhat.com/browse/PLANNER-2362


rsynek:
@ge0ffrey the Kafka integration is not a part of this PR. Shall we wait for it before merging?

kafka-quarkus-call-center/src/main/java/org/acme/callcenter/data/CallCenterDataGenerator.java
```diff
@@ -0,0 +1,86 @@
...
+import org.acme.callcenter.domain.CallCenter;
+import org.acme.callcenter.domain.Skill;
+
+@ApplicationScoped
+public class CallCenterDataGenerator {
+
+    private static final AtomicLong ID = new AtomicLong(1L);
+
+    private static final Agent[] AGENTS = new Agent[] {
+            new Agent(nextId(), "Ann", buildSkillSet(Skill.ENGLISH, Skill.LIFE_INSURANCE)),
```
triceo: If this is the only place where you actually use `nextId()`, why not simply use a constant?
Seems like too much of a hassle, just to generate a short fixed sequence of numbers.

kafka-quarkus-call-center/src/main/java/org/acme/callcenter/data/CallCenterDataGenerator.java
```diff
@@ -0,0 +1,86 @@
...
+        return new CallCenter(EnumSet.allOf(Skill.class), Arrays.asList(AGENTS), new ArrayList<>());
+    }
+
+    public Call generateCall(int durationSeconds) {
+        return new Call(nextId(), generatePhoneNumber(), buildSkillSet(pickRandomLanguageSkill(), pickRandomProductSkill()),
+                durationSeconds);
+    }
+
+    private Skill pickRandomProductSkill() {
+        return PRODUCT_SKILLS[ThreadLocalRandom.current().nextInt(3)];
```
triceo: `3` as in `PRODUCT_SKILLS.length`? Maybe it should just say that.

kafka-quarkus-call-center/src/main/java/org/acme/callcenter/data/CallCenterDataGenerator.java
```diff
@@ -0,0 +1,86 @@
...
+    }
+
+    private Skill pickRandomLanguageSkill() {
+        return LANGUAGE_SKILLS[ThreadLocalRandom.current().nextInt(3)];
```
triceo: Dtto.

kafka-quarkus-call-center/src/main/java/org/acme/callcenter/data/CallCenterDataGenerator.java
```diff
@@ -0,0 +1,86 @@
...
+    }
+
+    private String generatePhoneNumber() {
+        Random random = ThreadLocalRandom.current();
```
triceo: Any reason why the randomness needs to be thread-local? Why isn't a good old-fashioned randomness good enough?

kafka-quarkus-call-center/src/main/java/org/acme/callcenter/domain/Call.java
```diff
@@ -0,0 +1,170 @@
...
+        for (Skill skill : getRequiredSkills()) {
+            if (!agent.getSkills().contains(skill)) {
+                count++;
+            }
+        }
+        return count;
+    }
+
+    @Override
+    public Integer getEndTime() {
```
triceo: If it were me, I'd use the `TimeGrain` pattern.
I do not like naked ints, they do not mean anything and they don't look well in constraint streams for that reason.

kafka-quarkus-call-center/src/main/java/org/acme/callcenter/domain/Call.java
```diff
@@ -0,0 +1,170 @@
...
+
+    public PreviousCallOrAgent getPreviousCallOrAgent() {
+        return previousCallOrAgent;
+    }
+
+    public Agent getAgent() {
+        return agent;
+    }
+
+    public Integer getEstimatedWaiting() {
```
triceo: Duration? See reasoning above.

kafka-quarkus-call-center/src/main/java/org/acme/callcenter/domain/Skill.java
```diff
@@ -0,0 +1,24 @@
+package org.acme.callcenter.domain;
+
+import com.fasterxml.jackson.annotation.JsonValue;
+
+public enum Skill {
+
+    ENGLISH("EN"),
+    GERMAN("DE"),
+    SPANISH("ES"),
```
triceo: Why no Czech? :-)

triceo (commented):
I have added some comments.
That said, Geoffrey should still be the primary reviewer of this work.

kafka-quarkus-call-center/src/main/java/org/acme/callcenter/data/CallCenterDataGenerator.java
rsynek: Definitely, good catch!

kafka-quarkus-call-center/src/main/java/org/acme/callcenter/data/CallCenterDataGenerator.java
rsynek: No really; below there is a method generating new calls on-demand. This fixed declaration just reuses the same ID generator.

kafka-quarkus-call-center/src/main/java/org/acme/callcenter/domain/Call.java
rsynek: I don't think the `TimeGrain` pattern is the best choice here, as we are talking about seconds, which implies quite a lot `TimeGrain` instances. Also, as time moves forward, we would have to keep removing the old `TimeGrain` instances and adding new ones.
On the other hand, the `Integer` probably can be replaced by `Duration` to add more meaning to it.

kafka-quarkus-call-center/src/main/java/org/acme/callcenter/data/CallCenterDataGenerator.java
rsynek: It may be good enough, but this particular method is accessed from multiple threads (see the `SimilationService`), so the `ThreadLocalRandom` should be a better fit as it doesn't suffer from thread congestion.

kafka-quarkus-call-center/src/main/java/org/acme/callcenter/domain/Call.java
```diff
@@ -0,0 +1,169 @@
...
+        for (Skill skill : getRequiredSkills()) {
+            if (!agent.getSkills().contains(skill)) {
+                count++;
+            }
+        }
+        return count;
+    }
+
+    @Override
+    public Duration getEndTime() {
```
triceo: I'm not sure this is better. `Duration` is not `endTime`, these are two different concepts.

kafka-quarkus-call-center/pom.xml
```diff
@@ -0,0 +1,200 @@
...
+      <artifactId>font-awesome</artifactId>
+      <version>5.15.1</version>
+      <scope>runtime</scope>
+    </dependency>
+    <dependency>
+      <groupId>org.webjars.npm</groupId>
+      <artifactId>js-joda</artifactId>
+      <version>1.11.0</version>
+      <scope>runtime</scope>
+    </dependency>
```
ge0ffrey: Please create a jira to use the same consistent the JavaScript timelib in all quickstarts.
In practice, given that we're ally very used to java.time (~ joda)'s API and given that momentjs sucks (not immutable), I currently also favor going for js-joda, but that decision hasn't been made yet. IIRC there was another good one too.

joda-js is BSD license, so that's fine.
Just a pity it's there's no "plain webjar" only "npm webjar" for it. No idea if that actually matters.

@Christopher-Chianelli would you be ok with facilitating the "which js time lib to use" and doing that cleanup work to get all quickstarts to use it?

kafka-quarkus-call-center/src/main/resources/META-INF/resources/app.js
```diff
@@ -0,0 +1,288 @@
...
+        skillRow.append($(`
+            <div class="col-xs-1 card" style="background-color:${color};margin:2px;padding:2px">
+                <span style="font-size:0.8em">${skill}</span>
+            </div>`)
+        );
+    });
+}
+
+function solve() {
+  $.post("/call-center/solve", function () {
```
yurloc: Consider using the [Fetch API](https://developer.mozilla.org/en-US/docs/Web/API/Fetch_API). It's a new API for making HTTP requests and it's supported by all major browsers.

See https://www.turbo360.co/tip/fetch-vs-ajax-vs-xhr-vs-axios-i9azsx for comparison.

I think we should standardize on APIs that are supported by browsers and do not need additional dependencies.

kafka-quarkus-call-center/src/main/java/org/acme/callcenter/data/CallCenterDataGenerator.java
```diff
@@ -0,0 +1,86 @@
...
+    private Skill pickRandomProductSkill() {
+        return PRODUCT_SKILLS[ThreadLocalRandom.current().nextInt(PRODUCT_SKILLS.length)];
+    }
+
+    private Skill pickRandomLanguageSkill() {
+        return LANGUAGE_SKILLS[ThreadLocalRandom.current().nextInt(PRODUCT_SKILLS.length)];
+    }
+
+    private String generatePhoneNumber() {
+        Random random = ThreadLocalRandom.current();
```
ge0ffrey: For the DemoDataGenarator in other examples, we use a **seeded** single Random instance to generate the data. This is very intentional, for live demos: it's always the same data (although if you're creating new data on the click of the button, at least it's the same type of call for the same skill set, even though it might come a bit later on sooner depending on the click). _That predictable demo data is worth its weight in gold_: some random demo data, which is correct, doesn't look good and makes the demo look like it fails even though it doesn't. For example, presume you have 10 employees and 1 employee with the skill French, then in a live demo, you randomly get 10 calls in French in a row. That's no fun to demo and explain to the audience.

ge0ffrey: Given the multi-threaded access on this method (which the other quickstarts don't have to worry about):

Proposal A) Status quo

Proposal B) Create only one Random instance seeded (typically we use 37 as a seed, but if that gives us 10 french calls in a row, we tweak it). Make this method `synchronized`. If need be make both the field and method static.

kafka-quarkus-call-center/src/main/java/org/acme/callcenter/data/CallCenterDataGenerator.java
```diff
@@ -0,0 +1,86 @@
...
+
+import org.acme.callcenter.domain.Agent;
+import org.acme.callcenter.domain.Call;
+import org.acme.callcenter.domain.CallCenter;
+import org.acme.callcenter.domain.Skill;
+
+@ApplicationScoped
+public class CallCenterDataGenerator {
+
+    private static final AtomicLong ID = new AtomicLong(1L);
```
ge0ffrey: (soft)
Typically we call this NEXT_ID because it already has the value of the next id between 2 calls and to avoid confusing that it is the ID of this class (which it is not).
Also, typically id's start at 0.

kafka-quarkus-call-center/src/main/java/org/acme/callcenter/data/CallCenterDataGenerator.java
```diff
@@ -0,0 +1,86 @@
...
+
+import javax.enterprise.context.ApplicationScoped;
+
+import org.acme.callcenter.domain.Agent;
+import org.acme.callcenter.domain.Call;
+import org.acme.callcenter.domain.CallCenter;
+import org.acme.callcenter.domain.Skill;
+
+@ApplicationScoped
+public class CallCenterDataGenerator {
```
ge0ffrey: (separate issue): In other quickstarts with call it DemoDataGenerator, but that is becoming difficult because now there are 5+ classes named like that. We should rename them as is done here.

kafka-quarkus-call-center/src/main/java/org/acme/callcenter/domain/AbstractIdentifiable.java
```diff
@@ -0,0 +1,37 @@
...
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.acme.callcenter.domain;
+
+import org.optaplanner.core.api.domain.lookup.PlanningId;
+
+public abstract class AbstractIdentifiable {
```
ge0ffrey: In quickstarts we typically don't create such a common class, because it spreads the code more, reducing "vertical density" of the code.
Just having the ID on each domain class isn't that much more code. And easier to understand for the users, as it's one less class to click back and forth to.

kafka-quarkus-call-center/src/main/java/org/acme/callcenter/domain/Call.java
```diff
@@ -0,0 +1,169 @@
...
+            return 0;
+        }
+        int count = 0;
+
+        for (Skill skill : getRequiredSkills()) {
+            if (!agent.getSkills().contains(skill)) {
+                count++;
+            }
+        }
+        return count;
```
ge0ffrey: Maybe use a stream? `return requiredSkills.stream().filter(skill -> !agent.getSkills().contains(skill)).count();`
Or is that less readable for users?

kafka-quarkus-call-center/src/main/java/org/acme/callcenter/domain/Call.java
ge0ffrey: I agree, if it's called endTime should be an instance of Local(Date)Time

kafka-quarkus-call-center/src/main/resources/META-INF/resources/app.js
```diff
@@ -0,0 +1,288 @@
...
+function formatDuration(duration) {
+    const Duration = JSJoda.Duration;
+    const hours = Math.floor(duration.seconds() / 3600);
+    const minutes = Math.floor((duration.seconds() % 3600) / 60);
+    const seconds = duration.seconds() % 60;
+    var formattedDuration = '';
+    if (hours > 0) {
+        formattedDuration += hours + 'h:';
+    }
+    formattedDuration += minutes + 'm:' + seconds + 's';
```
yurloc: Any particular reason for this unique, hybrid time format? Could we go with a standard one like `5m 30s` or `05:30`?

kafka-quarkus-call-center/src/main/resources/META-INF/resources/app.js
```diff
@@ -0,0 +1,288 @@
...
+        const waiting = formatDuration(Duration.between(startedTime, LocalTime.now()));
+        const estimatedWaiting = formatDuration(Duration.ofSeconds(Math.floor(call.estimatedWaiting)));
+        $(`<span style="font-size:0.8em">Waiting: ${waiting}</span><br/>`).appendTo(callCardContainer);
+        $(`<span style="font-size:0.8em">Estimated waiting: ${estimatedWaiting}</span>`)
+            .appendTo(callCardContainer);
+    }
+}
+
+function formatDuration(duration) {
+    const Duration = JSJoda.Duration;
```
yurloc: Unused variable.

kafka-quarkus-call-center/src/main/resources/META-INF/resources/app.js
```diff
@@ -0,0 +1,288 @@
+const skillToColorMap = new Map([
+    ['EN', '#edd400'],
```
yurloc: We use 2-space indentation in JavaScript.

kafka-quarkus-call-center/src/test/java/org/acme/callcenter/solver/SolverServiceTest.java
```diff
@@ -0,0 +1,156 @@
...
+    @Test
+    @Timeout(60)
+    void prolongCall() {
+        CallCenter inputProblem = dataGenerator.generateCallCenter();
+        Call call1 = new Call(1L, "123-456-7891", Skill.ENGLISH, Skill.CAR_INSURANCE);
+        Call call2 = new Call(2L, "123-456-7892", Skill.ENGLISH, Skill.CAR_INSURANCE);
+        inputProblem.getCalls().add(call1);
+        inputProblem.getCalls().add(call2);
+
+        CallCenter bestSolution  = solve(inputProblem, () -> {
```
yurloc: Extra space.

kafka-quarkus-call-center/src/main/java/org/acme/callcenter/solver/ResponseTimeUpdatingVariableListener.java
```diff
@@ -0,0 +1,80 @@
...
+public class ResponseTimeUpdatingVariableListener implements VariableListener<CallCenter, Call> {
+
+    @Override
+    public void beforeEntityAdded(ScoreDirector<CallCenter> scoreDirector, Call call) {
+
+    }
+
+    @Override
+    public void afterEntityAdded(ScoreDirector<CallCenter> scoreDirector, Call call) {
+        updateResponseTime(scoreDirector,call);
```
yurloc: Missing space.

kafka-quarkus-call-center/src/main/java/org/acme/callcenter/service/SimulationService.java
```diff
@@ -0,0 +1,153 @@
...
+        running.set(false);
+        if (addNewCallScheduledFuture != null) {
+            addNewCallScheduledFuture.cancel(true);
+            addNewCallScheduledFuture = null;
+        }
+    }
+
+    /**
+     * Cancels the scheduled end of a call in progress and schedules a new end, postponed by a minute.
+     * */
```
yurloc: Extra `* `. Javadoc comments end with `*/` only.

kafka-quarkus-call-center/src/main/resources/META-INF/resources/app.js
```diff
@@ -0,0 +1,288 @@
...
+    }
+  });
+
+  $('#solveButton').click(function () {
+    solve();
+  });
+  $('#stopSolvingButton').click(function () {
+    stopSolving();
+  });
+  $('#addCallSubmitButton').click(function () {
```
yurloc: There is no such button.

kafka-quarkus-call-center/src/main/resources/META-INF/resources/app.js
```diff
@@ -0,0 +1,288 @@
...
+
+$(document).ready(function () {
+  $.ajaxSetup({
+    headers: {
+      'Content-Type': 'application/json',
+      'Accept': 'application/json'
+    }
+  });
+
+  $('#solveButton').click(function () {
```
yurloc: Can be simplified to
```js
$('#solveButton').click(solve);
```

kafka-quarkus-call-center/src/main/resources/META-INF/resources/app.js
```diff
@@ -0,0 +1,288 @@
...
+  });
+  callFrequencyValue.html(callFrequencyRange.val());
+
+  callLengthRange.on('change', function() {
+      callLengthValue.html(callLengthRange.val());
+      restartSimulation(callFrequencyRange.val(), callLengthRange.val());
+  });
+  callLengthValue.html(callLengthRange.val());
+
+  jQuery.each(['put', 'delete'], function (i, method) {
```
yurloc: This could be eliminated by using `fetch()` instead of JQuery. See facility location's `app.js`.

kafka-quarkus-call-center/src/main/resources/META-INF/resources/app.js
```diff
@@ -0,0 +1,288 @@
...
+var autoRefreshIntervalId = null;
+var solving = false;
+
+function refresh() {
+    refreshSolvingButtons();
+    $.getJSON("/call-center", (callCenterData) => {
+
+        $("#score").text("Score: " + (callCenterData.score == null ? "?" : callCenterData.score));
+        printCallTable(callCenterData);
+    });
```
yurloc: Ideally, also get the solving status and refresh solving buttons based on the status retrieved from back end.

Otherwise, if the solver is running and you refresh the browser you'll get a green solve button. You can then start solving again which will cause an exception.

kafka-quarkus-call-center/src/main/java/org/acme/callcenter/domain/Call.java
```diff
@@ -0,0 +1,169 @@
...
+import com.fasterxml.jackson.annotation.JsonIgnore;
+
+@PlanningEntity
+public class Call extends PreviousCallOrAgent {
+
+    private String phoneNumber;
+    private Set<Skill> requiredSkills;
+    private Duration duration = Duration.ZERO;
+    private LocalTime started;
+    private LocalTime pickUpTime;
```
yurloc: Maybe improve naming consistency by either using `started`+`pickedUp` or `startTime`+`pickUpTime`.

kafka-quarkus-call-center/src/main/resources/META-INF/resources/app.js
```diff
@@ -0,0 +1,288 @@
...
+    addCall();
+  });
+
+  const callFrequencyRange = $('#callFrequencyRange');
+  const callFrequencyValue = $('#callFrequencyValue');
+  const callLengthRange = $('#callLengthRange');
+  const callLengthValue = $('#callLengthValue');
+
+  callFrequencyRange.on('change', function() {
+      callFrequencyValue.html(callFrequencyRange.val());
```
yurloc: Please update the `callFrequencyValue` on the `input` event so that the value is reflected as the slider is moved. Something like this:
```js

  callFrequencyRange.on('input', function() {
      callFrequencyValue.html(callFrequencyRange.val());
  });
```

kafka-quarkus-call-center/src/main/java/org/acme/callcenter/domain/CallCenter.java
```diff
@@ -0,0 +1,75 @@
...
+
+    @PlanningEntityCollectionProperty
+    @ValueRangeProvider(id = "callRange")
+    private List<Call> calls;
+
+    @PlanningScore
+    private HardSoftScore score;
+
+    public CallCenter() {
+        // Required by OptaPlanner.
```
ge0ffrey: (nitpick) In other quickstarts we do it like this:
```suggestion
    // No-arg constructor required for OptaPlanner
    public CallCenter() {
```

kafka-quarkus-call-center/src/main/java/org/acme/callcenter/domain/CallCenter.java
```diff
@@ -0,0 +1,75 @@
...
+import org.optaplanner.core.api.domain.solution.PlanningSolution;
+import org.optaplanner.core.api.domain.solution.ProblemFactCollectionProperty;
+import org.optaplanner.core.api.domain.valuerange.ValueRangeProvider;
+import org.optaplanner.core.api.score.buildin.hardsoft.HardSoftScore;
+
+@PlanningSolution
+public class CallCenter {
+
+    @ProblemFactCollectionProperty
+    private Set<Skill> skills;
```
ge0ffrey: (very soft) As usual, I prefer skillSet naming etc over skills, to make it clear it's a set and not a list, same for agents -> agentList, etc. Many optaplanner API's expose that naming convention (for example getConstraintMatchList() etc). So these changes create inconsistency, but not sure if that's important enough. If you care deeply about rather seeing "skills" etc (I know Lukas does, so that would make 2 persons), we can discuss later and leave this PR as is.

kafka-quarkus-call-center/src/main/java/org/acme/callcenter/rest/CallCenterResource.java
```diff
@@ -0,0 +1,75 @@
...
+import javax.ws.rs.Path;
+import javax.ws.rs.Produces;
+import javax.ws.rs.core.MediaType;
+
+import org.acme.callcenter.data.CallCenterDataGenerator;
+import org.acme.callcenter.domain.CallCenter;
+import org.acme.callcenter.service.SimulationService;
+import org.acme.callcenter.service.SolverService;
+
+@Path("/call-center")
```
ge0ffrey: In TimeTableResource we use camelCase for this too

```suggestion
@Path("/callCenter")
```

If we go non-camel case for URL's, then the other quickstarts should do the same.

kafka-quarkus-call-center/src/main/java/org/acme/callcenter/rest/SimulationParameters.java
```diff
@@ -0,0 +1,39 @@
...
+ */
+
+package org.acme.callcenter.rest;
+
+public class SimulationParameters {
+    private int frequency;
+    private int duration;
+
+    public SimulationParameters() {
+        // Required by Jackson.
```
ge0ffrey:
```suggestion
    // No-arg constructor required for Jackson
    public SimulationParameters() {
```

kafka-quarkus-call-center/src/main/java/org/acme/callcenter/rest/json/DateTimeModuleObjectMapperCustomizer.java
```diff
@@ -0,0 +1,33 @@
...
+import com.fasterxml.jackson.datatype.jsr310.JavaTimeModule;
+
+import io.quarkus.jackson.ObjectMapperCustomizer;
+
+@ApplicationScoped
+public class DateTimeModuleObjectMapperCustomizer implements ObjectMapperCustomizer {
+
+    @Override
+    public void customize(ObjectMapper objectMapper) {
+        objectMapper.registerModule(new JavaTimeModule());
```
ge0ffrey: (important)

-1 on this. Quarks by default adds the JavatimeModule to the ObjectMapper it creates, as well as OptaPlanner scoring stuff etc. We should not create our own ObjectMapper instances, but inject them instead.

kafka-quarkus-call-center/src/main/java/org/acme/callcenter/service/SimulationService.java
```diff
@@ -0,0 +1,153 @@
...
+    private SolverService solverService;
+    private CallCenterDataGenerator dataGenerator;
+    private ConcurrentMap<Long, CallInProgress> callsInProgress = new ConcurrentHashMap<>();
+    private AtomicBoolean running = new AtomicBoolean(false);
+
+    @Inject
+    public SimulationService(SolverService solverService, CallCenterDataGenerator dataGenerator) {
+        this.solverService = solverService;
+        this.dataGenerator = dataGenerator;
+
```
ge0ffrey: (nitpick) empty line unneeded

kafka-quarkus-call-center/src/main/java/org/acme/callcenter/service/SimulationService.java
```diff
@@ -0,0 +1,153 @@
...
+        newBestSolution.getCalls().forEach(call -> {
+            if (call.getPreviousCallOrAgent() != null && call.getPreviousCallOrAgent() instanceof Agent) {
+                if (!callsInProgress.containsKey(call.getId())) {
+                    // Schedule finishing a call by an agent.
+                    ScheduledFuture<?> existingCallScheduledFuture =
+                            scheduleCallEnd(call, call.getDuration().getSeconds(), TimeUnit.SECONDS);
+                    // Pick-up time needs to be set, as it hasn't been propagated to this best solution yet.
+                    call.setPickUpTime(LocalTime.now());
+                    callsInProgress.put(call.getId(), new CallInProgress(call, existingCallScheduledFuture));
+                }
```
ge0ffrey: Could there be a race condition here because the callsInProgress condition (which in in itself is thread safe) and the action do when that happens are not atomic together?

kafka-quarkus-call-center/src/main/java/org/acme/callcenter/service/SolverService.java
```diff
@@ -0,0 +1,132 @@
...
+import org.optaplanner.core.api.solver.Solver;
+import org.optaplanner.core.api.solver.SolverFactory;
+import org.optaplanner.core.api.solver.event.BestSolutionChangedEvent;
+
+@ApplicationScoped
+public class SolverService {
+
+    ManagedExecutor managedExecutor;
+
+    private Solver<CallCenter> solver;
```
ge0ffrey: Can we add a "// TODO replace by @Inject SolverManager" with a link to the SM.addPFC jira?

kafka-quarkus-call-center/src/main/resources/application.properties
```diff
@@ -0,0 +1,21 @@
...
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+#
+
+########################
+# OptaPlanner properties
+########################
+quarkus.optaplanner.solver.termination.spent-limit=30s
+quarkus.optaplanner.solver.daemon=true
```
ge0ffrey: Can we get this consistent with the other application.properties files? They include info on how to turn on multithreaded solving, logging and FULL_ASSERT

ge0ffrey (approved):
Looks good. Very impressive demo!

Seperate PR suggestion: each quickstart has a screenshot, see `build/quickstarts-showcase/src/main/resources/META-INF/resources/screenshot/quarkus-vaccination-scheduling-screenshot.png` and where's that used (README + Showcase).
This quickstart should be mentioned in the showcase too, even if it can't be run from the showcase (see how we mention the optawebs there, without an option to launch it).

kafka-quarkus-call-center/pom.xml
Christopher-Chianelli: @ge0ffrey sure; JIRA issue: https://issues.redhat.com/browse/PLANNER-2403

kafka-quarkus-call-center/src/main/resources/META-INF/resources/app.js
rsynek: Good catch, thanks.

kafka-quarkus-call-center/src/main/resources/META-INF/resources/app.js
rsynek: No, this was my fabrication after realizing there is no standard way of formatting the `Duration`. I will go with the `5m 30s` format.

kafka-quarkus-call-center/src/main/resources/META-INF/resources/app.js
rsynek: Thanks for a nice hint, I will look into this.

kafka-quarkus-call-center/src/main/java/org/acme/callcenter/solver/ResponseTimeUpdatingVariableListener.java
rsynek: Thanks. A pity we don't have the formatted plugin in Quickstarts (and won't have, as we intentionally don't provide a parent pom for individual quickstarts).

kafka-quarkus-call-center/src/main/java/org/acme/callcenter/domain/Call.java
rsynek: No, I think the stream reads slightly better.

kafka-quarkus-call-center/src/main/java/org/acme/callcenter/domain/CallCenter.java
rsynek: I would rather keep it the way it is, as there is no chance the comment gets lost or moved during any future refactoring.

kafka-quarkus-call-center/src/main/java/org/acme/callcenter/domain/CallCenter.java
rsynek: I would like to keep the name `skills`: there is no need to duplicate the information about the type. Luckily, Java is a strongly and statically typed language. On the contrary, the Clean Code book specifically recommends not to encode types into a variable name.

kafka-quarkus-call-center/src/main/java/org/acme/callcenter/rest/CallCenterResource.java
rsynek: I quite agree with the argument [here](https://restfulapi.net/resource-naming/) that using hyphens instead of the camelCase improves readability. Look at the URL of this page. It's `optaplanner-quickstarts` not `optaplannerQuickstarts`.

+1 to change to this convention in the other quickstarts, if we seek absolute consistency.

kafka-quarkus-call-center/src/main/java/org/acme/callcenter/rest/SimulationParameters.java
rsynek: See my answer above.

kafka-quarkus-call-center/src/main/java/org/acme/callcenter/rest/json/DateTimeModuleObjectMapperCustomizer.java
rsynek: Interesting, I was definitely not able to pass the Java Time types via REST without this customization. I cannot reproduce it anymore.

kafka-quarkus-call-center/src/main/java/org/acme/callcenter/data/CallCenterDataGenerator.java
rsynek: https://github.com/kiegroup/optaplanner-quickstarts/pull/101

build/quickstarts-showcase/src/main/resources/META-INF/resources/index.html
```diff
@@ -91,6 +91,21 @@
                     </div>
                 </div>
             </div>
+            <div class="col mb-4">
+                <div class="card">
+                    <div class="card-header">Call center</div>
+                    <img src="screenshot/quarkus-call-center-screenshot.png" class="card-img-top mt-3" alt="Screenshot"/>
+                    <div class="card-body">
+                        <p class="card-text">
+                            Assign incoming calls to agents based on required skills and minimize the customer waiting time.
```
triceo:
```suggestion
                            Assign incoming calls to agents based on required skills while minimizing customer wait times.
```

quarkus-call-center/README.adoc
```diff
@@ -0,0 +1,70 @@
+= Call center realtime scheduling (Java, Quarkus, Maven)
```
triceo:
```suggestion
= Call center real-time scheduling (Java, Quarkus, Maven)
```

If we want to be gramatically correct here.

quarkus-call-center/src/main/java/org/acme/callcenter/domain/Call.java
```diff
@@ -0,0 +1,167 @@
...
+        }
+
+        return (int) requiredSkills.stream()
+                .filter(skill -> !agent.getSkills().contains(skill))
+                .count();
+    }
+
+    @Override
+    public Duration getDurationTillPickUp() {
+        Duration endTime;
```
triceo: Nitpick: `endTime` is not a duration. Duration is a length of time, not a point in time.

triceo (approved):
LGTM. Minor suggestions.

quarkus-call-center/src/main/java/org/acme/callcenter/domain/Call.java
rsynek: I missed this occurrence, thanks.

PR #101, opened 2021-04-23
Rename DemoDataGenerator classes

rsynek:

### DIFF
--- a/quarkus-facility-location/src/main/java/org/acme/facilitylocation/bootstrap/FacilityLocationDataGenerator.java
+++ b/quarkus-facility-location/src/main/java/org/acme/facilitylocation/bootstrap/FacilityLocationDataGenerator.java
@@ -26,11 +26,11 @@ import org.acme.facilitylocation.persistence.FacilityLocationProblemRepository;
 import io.quarkus.runtime.StartupEvent;
 
 @ApplicationScoped
-public class DemoDataGenerator {
+public class FacilityLocationDataGenerator {
 
     private final FacilityLocationProblemRepository repository;
 
-    public DemoDataGenerator(FacilityLocationProblemRepository repository) {
+    public FacilityLocationDataGenerator(FacilityLocationProblemRepository repository) {
         this.repository = repository;
     }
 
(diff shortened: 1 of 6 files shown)


quarkus-facility-location/src/main/java/org/acme/facilitylocation/bootstrap/FacilityLocationDataGenerator.java
```diff
@@ -26,11 +26,11 @@
 import io.quarkus.runtime.StartupEvent;
 
 @ApplicationScoped
-public class DemoDataGenerator {
+public class FacilityLocationDataGenerator {
```
ge0ffrey, line 29: I guess it should be FacilityLocationProblemDataGenerator as it doesn't generate just locations?

ge0ffrey (approved):
Good to merge if the FLP one is good too

ge0ffrey:
When we sync the quarkus-school-timetabling example with the one in quarkus-quickstarts, we need to adjust the quarkus docs, because those say "DemoDataGenerator". Note that those quarkus docs are coming back into our docs too (and we'll just duplicat them there).

@emmurphy1 This might affect some of the already single-sourced docs too.

triceo:
> When we sync the quarkus-school-timetabling example with the one in quarkus-quickstarts, we need to adjust the quarkus docs, because those say "DemoDataGenerator". Note that those quarkus docs are coming back into our docs too (and we'll just duplicat them there).

This is exactly why it makes little sense to rename things just to rename things.
The fix is worse than the original problem - not only do you have to change references in many places (and potentially miss some, actually *creating* incosistency), you also *destroy what people already remember*. Everyone who ever blogged or made any kind of note about this will have their content outdated when this is merged. 

The hypothetical benefits from consistency in this case are, in my opinion, miniscule - especially since *the quickstarts were already consistent*, you just didn't like the name.

I implore you - stop doing this, @ge0ffrey . It helps no one, and it possibly hurts many.

ge0ffrey:
@triceo Some renames are worth their fallout. Some restructures are worth their fallout. Like splitting off optaplanner-quickstart (which was your proposal IIRC). So I won't just stop polishing such things, especially in quickstarts/examples which need to be crystal clear for the getting started experience. This one is arguable, I don't feel strongly about it - happy to drop this request. We didn't realize the impact all the way into the quarkus docs when we noticed the downside of having so many classes with the same name.

@rsynek Wdyt? is this rename still worth the effort? Or shall we drop it? your call.

rsynek:
It's true that we might impact some blog posts (although in our blog posts we don't mention this class) and as nobody feels strongly about merging this PR, I am closing it.

If we want to keep every potential blog post compatible, shall we only fix bugs after any quickstart makes it into the `stable` branch?

ge0ffrey:
We don't want to keep blog posts compatible. That is a non-goal.

We want to keep the docs compatible, including those in quarkus.io, which's master needs to be moved to optaplanner-docs anyway.

PR #115, opened 2021-05-31
Upgrade Quarkus in Kotlin quickstart

rsynek:
The Kotlin module was omitted during the last upgrade since it already had a different Quarkus version compared to the other quickstarts.


rsynek:
Merging without review as this PR fixes the build.

PR #125, opened 2021-06-23
Upgrade Quarkus to 2.0.0.Final

rsynek:

### DIFF
--- a/kotlin-quarkus-school-timetabling/pom.xml
+++ b/kotlin-quarkus-school-timetabling/pom.xml
@@ -11,8 +11,9 @@
     <maven.compiler.release>11</maven.compiler.release>
     <project.build.sourceEncoding>UTF-8</project.build.sourceEncoding>
 
-    <version.io.quarkus>1.13.4.Final</version.io.quarkus>
-    <version.kotlin>1.3.72</version.kotlin>
+    <version.io.quarkus>2.0.0.Final</version.io.quarkus>
+    <!-- The version 1.4.32 Quarkus uses does not work with Java 16. -->
+    <version.kotlin>1.5.10</version.kotlin>
     <version.org.optaplanner>8.9.0-SNAPSHOT</version.org.optaplanner>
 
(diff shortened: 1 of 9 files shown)


kotlin-quarkus-school-timetabling/pom.xml
```diff
@@ -11,8 +11,9 @@
     <maven.compiler.release>11</maven.compiler.release>
     <project.build.sourceEncoding>UTF-8</project.build.sourceEncoding>
 
-    <version.io.quarkus>1.13.4.Final</version.io.quarkus>
-    <version.kotlin>1.3.72</version.kotlin>
+    <version.io.quarkus>2.0.0.Final</version.io.quarkus>
+    <!-- The version 1.4.32 Quarkus uses does not work with Java 16. -->
```
ge0ffrey: I'd remove this comment in the code. It's intresting for the PR, but not for the long term code

spring-boot-school-timetabling/pom.xml
```diff
@@ -15,7 +15,8 @@
     <version.org.springframework.boot>2.4.5</version.org.springframework.boot>
 
     <version.compiler.plugin>3.8.1</version.compiler.plugin>
-    <version.surefire.plugin>2.22.2</version.surefire.plugin>
+    <!-- The version 2.22.2 fails when the tests run from the top-level project. -->
```
ge0ffrey: Same principle: I 'd remove this comment (that an older version fails is not suprising for any of the dep versions)

Are all the quickstarts on the same surefire version now? We should get them all on the same surefire version.

ge0ffrey (approved):
LGTM if comments are removed
 and if surefire is made consistent (which can be in a separate PR later)

spring-boot-school-timetabling/pom.xml
rsynek: Not yet; I would prefer doing it in a separate PR, as there is already quite some urgency on merging this one.

PR #142, opened 2021-07-21
PLANNER-2437 Enable changing Quarkus BOM GAV

rsynek:
### JIRA
https://issues.redhat.com/browse/PLANNER-2437

### DIFF
--- a/build/quickstarts-showcase/pom.xml
+++ b/build/quickstarts-showcase/pom.xml
@@ -13,6 +13,10 @@
   <name>OptaPlanner Quickstarts Showcase</name>
 
   <properties>
+    <quarkus.platform.artifact-id>quarkus-bom</quarkus.platform.artifact-id>
+    <quarkus.platform.group-id>io.quarkus</quarkus.platform.group-id>
+    <quarkus.platform.version>${version.io.quarkus}</quarkus.platform.version>
+
     <java.module.name>org.optaplanner.quickstarts.showcase</java.module.name>
     <!-- Workaround for https://github.com/quarkusio/quarkus/issues/15817. -->
     <version.compiler.plugin>3.8.1</version.compiler.plugin>
@@ -21,9 +25,9 @@
   <dependencyManagement>
     <dependencies>
       <dependency>
-        <groupId>io.quarkus</groupId>
-        <artifactId>quarkus-bom</artifactId>
-        <version>${version.io.quarkus}</version>
+        <groupId>${quarkus.platform.group-id}</groupId>
+        <artifactId>${quarkus.platform.artifact-id}</artifactId>
+        <version>${quarkus.platform.version}</version>
         <type>pom</type>
         <scope>import</scope>
       </dependency>
--- a/technology/java-activemq-quarkus/pom.xml
+++ b/technology/java-activemq-quarkus/pom.xml
@@ -10,6 +10,10 @@
   <packaging>pom</packaging>
 
   <properties>
+    <quarkus.platform.artifact-id>quarkus-bom</quarkus.platform.artifact-id>
+    <quarkus.platform.group-id>io.quarkus</quarkus.platform.group-id>
+    <quarkus.platform.version>${version.io.quarkus}</quarkus.platform.version>
+
     <version.org.optaplanner>8.10.0-SNAPSHOT</version.org.optaplanner>
     <version.io.quarkus>2.0.1.Final</version.io.quarkus>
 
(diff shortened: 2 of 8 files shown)


rsynek:
@mareknovotny how do we productize the `build.gradle` files? Shall I introduce the same properties here as well?
https://github.com/kiegroup/optaplanner-quickstarts/blob/development/use-cases/school-timetabling/build.gradle#L25

mareknovotny:
> @mareknovotny how do we productize the `build.gradle` files? Shall I introduce the same properties here as well?
> https://github.com/kiegroup/optaplanner-quickstarts/blob/development/use-cases/school-timetabling/build.gradle#L25

we don't productize gradle at all, product supports only Maven

mareknovotny (approved):
lgtm

build/quickstarts-showcase/pom.xml
```diff
@@ -13,6 +13,10 @@
   <name>OptaPlanner Quickstarts Showcase</name>
 
   <properties>
+    <quarkus.platform.artifact-id>quarkus-bom</quarkus.platform.artifact-id>
+    <quarkus.platform.group-id>io.quarkus</quarkus.platform.group-id>
```
ge0ffrey, line 17: Consistency: groupId is always before artifactIid (It's GAV, not AGV)

build/quickstarts-showcase/pom.xml
```diff
@@ -13,6 +13,10 @@
...
+    <quarkus.platform.version>${version.io.quarkus}</quarkus.platform.version>
```
ge0ffrey, line 18: Let's use version.io.quarkus like all our other version properties in all our other poms.

ge0ffrey (changes requested):
I have very mixed feelings about this groupId and artifactId property (not for the version property). It really complicates the getting started experience, both for the community and the product. Less code is better.
This is a step backwards for community.

@mareknovotny Would it be technically possible to overwrite the groupId and artifactId directly in the dependencyManagement during the productization? Can we create an track an issue for that?
If there's such an issue - and a commitment to prioritize fixing it - I can live with merging this PR as a temporarily workaround. Otherwise, we need to discuss.

build/quickstarts-showcase/pom.xml
mareknovotny: i agree, but here are the properties declared and we have JBoss standard to have alphabetic order, don't we?

build/quickstarts-showcase/pom.xml
mareknovotny: please be warned that it can be different version from version.io.quarkus as the platform will and can be released more times on the same quarkus version due other platform additions and rebuilds IMO.

mareknovotny:
> I have very mixed feelings about this groupId and artifactId property (not for the version property). It really complicates the getting started experience, both for the community and the product. Less code is better.
> This is a step backwards for community.
> 
> @mareknovotny Would it be technically possible to overwrite the groupId and artifactId directly in the dependencyManagement during the productization? Can we create an track an issue for that?
> If there's such an issue - and a commitment to prioritize fixing it - I can live with merging this PR as a temporarily workaround. Otherwise, we need to discuss.

What is the main problem here you don't like? the property declaration is set to a default and I think it is not a problem for getting started.
What is actually the problem is that you disable the flexible change of the platform for anybody. How could you explain to end users that we have different declarations or even sources of quickstarts are forked ?
We were just inspired by Quarkus quickstarts where we don't see any problems or complains ? or do we?

ge0ffrey:
> What is the main problem here you don't like? 

These changes increase the complexity of the build files for the community quickstarts. That effects the Getting Started experience for OptaPlanner users. I believe there are better ways to solve this issue and I would love to discuss that. Given that there apparently is no time for this meeting, I suggest we either fix this issue for the next product release and/or we make these changes only on the current product branch. To be crystal clear, without proper discussion in a meeting I am hard vetoing these changing for the `development` and `stable` branches.

> We were just inspired by Quarkus quickstarts where we don't see any problems or complains ?

There are plenty complains about that.

ge0ffrey:
We 've discussed this in a meeting and came to a conclusion that was ok for all participants:
- I will open a discussion with Max on the issue of the properties upstream in Quarkus.
- We will make the changes of the PR on the release branch (8.11.x) once its created, to not impact the product release.
- Right now, we will not make these changes on the development branch, but depending on how the upstream discussion goes, we might need to make them later or they will become obsolete.

PR #147, opened 2021-07-29
PLANNER-2458: OptaPlanner quick-starts - OrderPicking

wmedvede:


wmedvede:
Hi @ge0ffrey, would you mind have a look a the order picking problem model and constraints?

Thanks!

PR #164, opened 2021-08-25
[8.11] PLANNER-2437 Use Quarkus 2 platform BOM

rsynek:
As previously agreed, this change targets only the `8.11.x` branch, while the `development` branch remains untouched.

### JIRA
https://issues.redhat.com/browse/PLANNER-2437

### DIFF
--- a/technology/java-activemq-quarkus/pom.xml
+++ b/technology/java-activemq-quarkus/pom.xml
@@ -10,12 +10,14 @@
   <packaging>pom</packaging>
 
   <properties>
-    <version.org.optaplanner>8.11.0-SNAPSHOT</version.org.optaplanner>
-    <version.io.quarkus>2.1.2.Final</version.io.quarkus>
-
-    <project.build.sourceEncoding>UTF-8</project.build.sourceEncoding>
     <compiler-plugin.version>3.8.1</compiler-plugin.version>
     <maven.compiler.release>11</maven.compiler.release>
+    <project.build.sourceEncoding>UTF-8</project.build.sourceEncoding>
+
+    <quarkus.platform.artifact-id>quarkus-bom</quarkus.platform.artifact-id>
+    <quarkus.platform.group-id>io.quarkus.platform</quarkus.platform.group-id>
+    <quarkus.platform.version>2.1.2.Final</quarkus.platform.version>
+
     <surefire-plugin.version>3.0.0-M5</surefire-plugin.version>
   </properties>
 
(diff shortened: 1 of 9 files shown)


antncl:
Looks good to me, I would maybe add a section into the readme that describes how to override this, like here:
https://github.com/kiegroup/kogito-examples/tree/main#use-alternative-quarkus-platforms

rsynek:
> Looks good to me, I would maybe add a section into the readme that describes how to override this, like here:
> https://github.com/kiegroup/kogito-examples/tree/main#use-alternative-quarkus-platforms

Looking at the text in `kogito-examples`, I don't think it gives the best example. Although the `quarkus-universe-bom` is still part of the platform, it's recommended to use (multiple, if needed) fine-grained BOMs like described in https://quarkus.io/blog/quarkus-2x-platform-quarkiverse-registry/ .

technology/java-activemq-quarkus/pom.xml
```diff
@@ -28,18 +30,16 @@
   <dependencyManagement>
     <dependencies>
       <dependency>
-        <groupId>io.quarkus</groupId>
-        <!-- Alternatively, use <artifactId>quarkus-universe-bom</artifactId>
-         which includes both quarkus-bom and optaplanner-bom. -->
-        <artifactId>quarkus-bom</artifactId>
-        <version>${version.io.quarkus}</version>
+        <groupId>${quarkus.platform.group-id}</groupId>
+        <artifactId>quarkus-optaplanner-bom</artifactId>
```
rsynek, line 34: @AntStephenson With the Quarkus 2.2.0.Final, the platform should already use OptaPlanner build parent as an input to generate this `quarkus-optaplanner-bom`. This artifact should define also versions of dependencies, so in theory, we should not need the core `quarkus-bom`.

antncl: @rsynek When I tried on Kogito Examples to build with just the quarkus-kogito-bom the build failed, I guess this changed with 2.2.0.Final? if that's true I guess that's OK. There is also evolving discussions on that ticket I linked you to for Kogito too about how it should work.

antncl (approved):

technology/java-activemq-quarkus/pom.xml
rsynek: Not sure how far is Kogito with including the 3rd party dependencies into their BOM. Anyway, it should still work, just making the quarkus-bom unnecessary.

PR #141, opened 2021-07-20
PLANNER-2474 optaplanner CI

Ginxo:
### JIRA
https://issues.redhat.com/browse/PLANNER-2474

`https://raw.githubusercontent.com/kiegroup/kogito-pipelines/main/.ci/pull-request-config.yaml` is pointing to a specific branch, it should be dynamically calculated, this is a pending improvement on build-chain tool https://github.com/kiegroup/github-action-build-chain/issues/117

### Referenced pull requests
* https://github.com/kiegroup/optaplanner/pull/1440
* https://github.com/kiegroup/optaplanner-quickstarts/pull/141
* https://github.com/kiegroup/kogito-pipelines/pull/215
* https://github.com/kiegroup/optaweb-employee-rostering/pull/677
* https://github.com/kiegroup/optaweb-vehicle-routing/pull/547

PR #176, opened 2021-09-07
Fix GHA workflows

yurloc:
The error says:

> Invalid workflow file
> No steps defined in `steps` for the following jobs: concurrency

`concurrency` must be either at the top level or under a specific `job`, not under `jobs`.

https://docs.github.com/en/actions/reference/workflow-syntax-for-github-actions#concurrency

Fixup for #141.

### DIFF
--- a/.github/workflows/full-downstream.yml
+++ b/.github/workflows/full-downstream.yml
@@ -16,10 +16,10 @@ on:
       - '.ci/**'
 
 jobs:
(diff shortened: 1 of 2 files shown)


triceo (changes requested):
Please drop JDK 8. Otherwise looks good to me.

triceo (approved):

PR #184, opened 2021-09-24
Fix VRP SolverResourceTest

yurloc:
With the current termination configured in the test profile, solving
ends when the CH is done. This is too short for the test to be able to
reliably verify the solver is solving. So instead of that, just wait
until it stops solving.

### DIFF
--- a/.github/workflows/full-downstream.yml
+++ b/.github/workflows/full-downstream.yml
@@ -1,11 +1,11 @@
 
-name: Build Chain
+name: [FDB] Build Chain
 
 on:
   pull_request:
-    types: [opened, synchronize, reopened, ready_for_review]
+    types: [labeled]
     branches:
-      - development
+      - main
       - 8.*
     paths-ignore:
       - 'LICENSE*'
@@ -17,8 +17,9 @@ on:
 
 jobs:
   build-chain:
+    if: contains(github.event.pull_request.labels.*.name, 'run_fdb')
     concurrency:
-      group: pull_request-${{ github.head_ref }}
+      group: fdb-${{ github.head_ref }}
       cancel-in-progress: true
     strategy:
       matrix:
--- a/.github/workflows/full-downstream.yml
+++ b/.github/workflows/full-downstream.yml
@@ -5,7 +5,7 @@ on:
   pull_request:
     types: [labeled]
     branches:
-      - main
+      - development
       - 8.*
     paths-ignore:
       - 'LICENSE*'
--- a/.github/pull_request_template.md
+++ b/.github/pull_request_template.md
@@ -33,6 +33,7 @@ How to retest this PR or trigger a specific build:
 </summary>
 
 * for a <b>pull request build</b> please add comment: <b>Jenkins retest this</b>
+* for a <b>full downstream build</b> please add the label `run_fdb`
 * for a <b>specific pull request build</b> please add comment: <b>Jenkins (re)run [optaplanner-quickstarts] tests</b>
 * for a <b>Quarkus LTS check</b> please add comment: <b>Jenkins run LTS</b>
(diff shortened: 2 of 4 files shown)


radtriste:
jenkins retest this

yurloc:
Why is one of the jobs canceled?

Do we need two build chain workflows?

radtriste:
not sure what is happening on the build-chain ... @Ginxo ?

Ginxo:
@radtriste @yurloc build-chain has nothing to do with job cancellation, in any case is GHA configuration, but anyway I don't see any reason on workflow configuration for the jobs to be cancelled. So... I've restarted them and the jobs are not cancelled anymore... it seems to be a GHA exception here... let's wait for jobs completion.

yurloc:
@Ginxo could it be caused by `cancel-in-progress: true`? https://docs.github.com/en/actions/learn-github-actions/workflow-syntax-for-github-actions#concurrency

Ginxo:
@yurloc I didn't check the right files.
So it is obviously due to 
https://github.com/kiegroup/optaplanner-quickstarts/blob/ca673db4dd1a1ed36117d74783675a00b29ad90a/.github/workflows/pull_request.yml#L20
and
https://github.com/kiegroup/optaplanner-quickstarts/blob/ca673db4dd1a1ed36117d74783675a00b29ad90a/.github/workflows/full-downstream.yml#L20
So fdb and pr jobs are colliding... Pushing a commit to fix this and also making FDB job to be run by label (as we do in optaplanner)

Ginxo:
@yurloc @radtriste look:
https://github.com/yurloc/optaplanner-quickstarts/commit/46c60928bf5588fc7792d6cd939f48efea81abe0#diff-dade13b21248589cfd115a1a19122645302300751c49d3a67d176a5349e36330
https://github.com/yurloc/optaplanner-quickstarts/commit/63ba90bb8c08807c0dd1d99131dda6ec4fc93ef4#diff-b2496e80299b8c3150b1944450bd81c622e04e13d15c411d291db0927d75fd6b

@yurloc can you please rebase

yurloc:
@Ginxo why is there such a collision? Do we really need 2 separate workflow files both running FDB jobs that collide with each other?

.github/workflows/full-downstream.yml
```diff
@@ -1,11 +1,11 @@
...
-name: Build Chain
+name: [FDB] Build Chain
 
 on:
   pull_request:
-    types: [opened, synchronize, reopened, ready_for_review]
+    types: [labeled]
     branches:
-      - development
+      - main
```
yurloc: @Ginxo There is no `main` branch in this repo.

Ginxo: you areright @yurloc

Ginxo:
@yurloc 

- one of the jobs (pull_request.yml) is running a pull request job (PR's project plus its upstream dependencies).
- the other job (full-downstream.yml) is running a full downstream (PR's project plus its upstream + downstream dependencies)

they are colliding due to they are using same `concurrency.group` key, which is wrong. My proposal is obviously replace the one from full-downstream.yml and additionally be able to run it on demand (by adding `run_fdb` label to the PR)

radtriste:
I think that first part of `os_java_ci` yaml file and `pull-request` yaml file are checking the same ...
we would just have check for JDK 16 (soon 17?) in os_java_ci and maybe rename that file.
wdyt ?
and maybe also use the build-chain with JDK 16 ?

triceo:
Folks, I introduced another PR:
https://github.com/kiegroup/optaplanner-quickstarts/pull/186

This one removes os_java_ci, as we do not need that one. (pull_request does the same thing, but better.)
It also adds JDK 17.
I am running in the cancellation issue though.

triceo (approved):
No concerns; will merge to get the CI fix.

PR #186, opened 2021-09-29
PLANNER-2526 Introduce JDK 17

triceo:
Removing `os_java_ci`, as it is now the same thing as `pull_request`, only worse.
OptaPlanner repo did the same.


Ginxo:
+1 @triceo 
I would take profit of this PR to upgrade `kiegroup/github-action-build-chain@v2.6.5` from https://github.com/triceo/optaplanner-quickstarts/blob/aeae81f378d97c7b312c9eb5338ca21b8bf072ae/.github/workflows/pull_request.yml#L47 to `kiegroup/github-action-build-chain@v2.6.8` this way we will take latest changes (mainly the `print execution plan` functionality https://github.com/kiegroup/github-action-build-chain/issues/167)
@triceo can you do so? thanks!

Ginxo (approved):
LGTM, thanks @triceo nice contribution!

.github/workflows/pull_request.yml
```diff
@@ -18,12 +18,12 @@ on:
 jobs:
   build-chain:
     concurrency:
-      group: pull_request-${{ github.head_ref }}
+      group: pull_request-${{ github.event_name }}-${{ github.head_ref }}-${{ matrix.os }}-${{ matrix.java-version }}
```
yurloc, line 21: @triceo 

Does this mean that a job will only be cancelled if:
- it's a pull request
- from the same branch
- it's the same OS x Java combination?
So basically cancellation will only appear if I update my PR and there are still unfinished jobs pending from the previous PR update, is that correct?

triceo: Yes. As far as I know, this was the goal.

Ginxo: it means a running job will be cancelled in case a new pull_request job for the same PR-branch/OS/java-version is triggered. The real thing is, in case one of the matrix jobs is cancelled the whole job will be cancelled. 

Let's suppose we have this situation:

- PR ubuntu java 11 done
- PR ubuntu java 17 done
- PR windows java 11 done
- PR windows java 17 running

and a new commit is pushed to the PR the 4th check_run (windows+17) will be cancelled (the rest will be respected as they are) and a new set of check_run will be triggered. Then group including OS and java-version does not make much sense but they don't disturb either... In order to simplify things we can keep group as `pull_request-${{ github.event_name }}-${{ github.head_ref }}` and the final result will be the same. Up to you guys

triceo: Hmm, good point, I might as well remove the os/jvm from there and the end result will be exactly the same.

yurloc (approved):
Thanks for the explanation. LGTM.

.github/workflows/pull_request.yml
triceo, line 21:
```suggestion
      group: pull_request-${{ github.event_name }}-${{ github.head_ref }}
```

.github/workflows/pull_request.yml
```diff
@@ -18,12 +18,12 @@ on:
 jobs:
   build-chain:
     concurrency:
-      group: pull_request-${{ github.head_ref }}
+      group: pull_request-${{ github.event_name }}-${{ github.head_ref }}
```
triceo:
```suggestion
      group: pull_request-${{ github.event_name }}-${{ github.head_ref }}-${{ matrix.os }}-${{ matrix.java-version }}
```

PR #187, opened 2021-09-30
PLANNER-2458: OptaPlanner quick-starts - OrderPicking

wmedvede:


wmedvede:
Hi  @ge0ffrey , would you mind review this PR?

use-cases/school-timetabling/src/main/resources/META-INF/resources/index.html
```diff
@@ -30,7 +30,7 @@
         </a>
     </nav>
     <div class="sticky-top d-flex justify-content-center align-items-center" aria-live="polite" aria-atomic="true">
-        <div id="notificationPanel" style="position: absolute; top: .5rem;"></div>
+        <div id="notificationPanel" style="side: absolute; top: .5rem;"></div>
```
yurloc: Looks like an unintended side effect of refactoring.

wmedvede: damn, good catch.

wmedvede: fixed

use-cases/order-picking/src/main/java/org/acme/orderpicking/bootstrap/DemoDataGenerator.java
```diff
@@ -0,0 +1,122 @@
...
+            TrolleyStep trolleyStep = new TrolleyStep(item);
+            steps.add(trolleyStep);
+        }
+        return steps;
+    }
+
+    public void validateBucketCapacity(int bucketCapacity) {
+        if (bucketCapacity < ProductDB.getMaxProductSize()) {
+            throw new IllegalArgumentException("The selected bucketCapacity: " + bucketCapacity + ", is lower than the maximum product size: " + ProductDB.getMaxProductSize() + "." +
+                    " However for a matter of simplicity the problem was simplified on the assumption that products can always feet on a trolley bucket." +
```
yurloc:
```suggestion
                    " However for a matter of simplicity the problem was simplified on the assumption that products can always fit in a trolley bucket." +
```

use-cases/order-picking/src/main/java/org/acme/orderpicking/bootstrap/DemoDataGenerator.java
```diff
@@ -0,0 +1,122 @@
...
+        }
+        return orderList;
+    }
+
+    private List<Product> buildProducts() {
+        return ProductDB.getProducts().stream()
+                .map(productRecord -> {
+                    List<String> shelvingIds = ProductDB.getShelvings(productRecord.getFamily());
+                    int shelvingIndex = random.nextInt(shelvingIds.size());
+                    Shelving.Side shelvingSide = Shelving.Side.values()[random.nextInt(2)];
```
yurloc:
```suggestion
                    Shelving.Side shelvingSide = Shelving.Side.values()[random.nextInt(Shelving.Side.values().length)];
```
Let's avoid magical constants.

use-cases/order-picking/src/main/resources/META-INF/resources/warehouse-api.js
```diff
@@ -0,0 +1,183 @@
...
+const SHELVING_STROKE_STYLE = 'green';
+const SHELVING_LINE_JOIN = 'bevel';
+
+const TROLLEY_PATH_LINE_WIDTH = 5;
+const TROLLEY_PATH_LINE_JOIN = 'round';
+
+const SHELVINGS_MAP = new Map();
+
+function shelvingId(i, j) {
+    return '(' + i + ',' + j + ')';
```
yurloc, line 41: I would prefer if the shelvings were labeled in a bit more human-friendly way, something like `A1, A2, ..., B1, ...`. If someone does a demo of this quickstart and needs to explain the solution in words, it will make their life easier, I think.

use-cases/order-picking/src/main/java/org/acme/orderpicking/bootstrap/DemoDataGenerator.java
```diff
@@ -0,0 +1,122 @@
...
+        for (int orderNumber = 1; orderNumber <= size; orderNumber++) {
+            int orderItemsSize = ORDER_ITEMS_SIZE_MINIMUM + random.nextInt(products.size() - ORDER_ITEMS_SIZE_MINIMUM);
+            List<OrderItem> orderItems = new ArrayList<>();
+            Set<String> orderProducts = new HashSet<>();
+            order = new Order(Integer.toString(orderNumber), orderItems);
+            int itemNumber = 1;
+            for (int i = 0; i < orderItemsSize; i++) {
+                int productItemIndex = random.nextInt(products.size());
+                Product product = products.get(productItemIndex);
+                if (!orderProducts.contains(product.getId())) {
```
yurloc, line 98: Is it OK that the number of items in an order is often lower than `orderItemsSize`? I find it a bit confusing but I can imagine it can be tolerated for the sake of simplicity.

use-cases/order-picking/src/main/java/org/acme/orderpicking/bootstrap/ProductDB.java
```diff
@@ -0,0 +1,196 @@
...
+        PRODUCTS.add(new ProductDBRecord(new Product(nextId(), "Mature Cheddar 400G", 10 * 9 * 5, null), ProductFamily.FRESH_FOOD));
+        PRODUCTS.add(new ProductDBRecord(new Product(nextId(), "Butter Packet", 12 * 5 * 5, null), ProductFamily.FRESH_FOOD));
+
+        PRODUCTS.add(new ProductDBRecord(new Product(nextId(), "Iceberg Lettuce Each", 2500, null), ProductFamily.FRUITS_AND_VEGETABLES));
+        PRODUCTS.add(new ProductDBRecord(new Product(nextId(), "Carrots 1Kg", 1000, null), ProductFamily.FRUITS_AND_VEGETABLES));
+        PRODUCTS.add(new ProductDBRecord(new Product(nextId(), "Organic Fair Trade Bananas 5 Pack", 1800, null), ProductFamily.FRUITS_AND_VEGETABLES));
+        PRODUCTS.add(new ProductDBRecord(new Product(nextId(), "Gala Apple Minimum 5 Pack", 25 * 20 * 10, null), ProductFamily.FRUITS_AND_VEGETABLES));
+        PRODUCTS.add(new ProductDBRecord(new Product(nextId(), "Orange Bag 3kg", 29 * 20 * 15, null), ProductFamily.FRUITS_AND_VEGETABLES));
+
+        PRODUCTS.add(new ProductDBRecord(new Product(nextId(), "Fairy Non Biological Laundry Liquid 4.55L ", 5000, null), ProductFamily.HOUSE_CLEANING));
```
yurloc:
```suggestion
        PRODUCTS.add(new ProductDBRecord(new Product(nextId(), "Fairy Non Biological Laundry Liquid 4.55L", 5000, null), ProductFamily.HOUSE_CLEANING));
```
removing trailing space

use-cases/order-picking/src/main/java/org/acme/orderpicking/bootstrap/ProductDB.java
```diff
@@ -0,0 +1,196 @@
...
+
+        PRODUCTS.add(new ProductDBRecord(new Product(nextId(), "Tea Bags 240 packet", 2 * 6 * 15, null), ProductFamily.GENERAL_FOOD));
+        PRODUCTS.add(new ProductDBRecord(new Product(nextId(), "Tomato Soup Can", 10 * 10 * 10, null), ProductFamily.GENERAL_FOOD));
+        PRODUCTS.add(new ProductDBRecord(new Product(nextId(), "Baked Beans in Tomato Sauce", 10 * 10 * 10, null), ProductFamily.GENERAL_FOOD));
+
+        PRODUCTS.add(new ProductDBRecord(new Product(nextId(), "Classic Mint Sauce", 8 * 10 * 8, null), ProductFamily.GENERAL_FOOD));
+        PRODUCTS.add(new ProductDBRecord(new Product(nextId(), "Raspberry Conserve", 8 * 10 * 8, null), ProductFamily.GENERAL_FOOD));
+        PRODUCTS.add(new ProductDBRecord(new Product(nextId(), "Orange Fine Shred Marmalade", 7 * 8 * 7, null), ProductFamily.GENERAL_FOOD));
+
+        PRODUCTS.add(new ProductDBRecord(new Product(nextId(), "Free Range Eggs 6 Pack ", 15 * 10 * 8, null), ProductFamily.FRESH_FOOD));
```
yurloc:
```suggestion
        PRODUCTS.add(new ProductDBRecord(new Product(nextId(), "Free Range Eggs 6 Pack", 15 * 10 * 8, null), ProductFamily.FRESH_FOOD));
```
removing trailing space

use-cases/order-picking/src/main/java/org/acme/orderpicking/bootstrap/ProductDB.java
```diff
@@ -0,0 +1,196 @@
...
+        PRODUCTS.add(new ProductDBRecord(new Product(nextId(), "Orange Bag 3kg", 29 * 20 * 15, null), ProductFamily.FRUITS_AND_VEGETABLES));
+
+        PRODUCTS.add(new ProductDBRecord(new Product(nextId(), "Fairy Non Biological Laundry Liquid 4.55L ", 5000, null), ProductFamily.HOUSE_CLEANING));
+        PRODUCTS.add(new ProductDBRecord(new Product(nextId(), "Toilet Tissue 8 Roll White", 50 * 20 * 20, null), ProductFamily.HOUSE_CLEANING));
+        PRODUCTS.add(new ProductDBRecord(new Product(nextId(), "Kitchen Roll 200 Sheets x 2", 30 * 30 * 15, null), ProductFamily.HOUSE_CLEANING));
+        PRODUCTS.add(new ProductDBRecord(new Product(nextId(), "Stainless Steel Cleaner 500Ml", 500, null), ProductFamily.HOUSE_CLEANING));
+        PRODUCTS.add(new ProductDBRecord(new Product(nextId(), "Antibacterial Surface Spray", 12 * 4 * 25, null), ProductFamily.HOUSE_CLEANING));
+
+        PRODUCTS.add(new ProductDBRecord(new Product(nextId(), "Beef Lean Steak Mince 500g", 500, null), ProductFamily.MEET_AND_FISH));
+        PRODUCTS.add(new ProductDBRecord(new Product(nextId(), "Smoked Salmon 120G ", 150, null), ProductFamily.MEET_AND_FISH));
```
yurloc:
```suggestion
        PRODUCTS.add(new ProductDBRecord(new Product(nextId(), "Smoked Salmon 120G", 150, null), ProductFamily.MEET_AND_FISH));
```
removing trailing space

use-cases/order-picking/src/main/java/org/acme/orderpicking/bootstrap/ProductDB.java
```diff
@@ -0,0 +1,196 @@
...
+        PRODUCTS.add(new ProductDBRecord(new Product(nextId(), "Steak Burgers 454G", 450, null), ProductFamily.MEET_AND_FISH));
+        PRODUCTS.add(new ProductDBRecord(new Product(nextId(), "Pork Cooked Ham 125G", 125, null), ProductFamily.MEET_AND_FISH));
+        PRODUCTS.add(new ProductDBRecord(new Product(nextId(), "Chicken Breast Fillets 300G", 300, null), ProductFamily.MEET_AND_FISH));
+
+        PRODUCTS.add(new ProductDBRecord(new Product(nextId(), "6 Milk Bricks Pack ", 22 * 16 * 21, null), ProductFamily.DRINKS));
```
yurloc:
```suggestion
        PRODUCTS.add(new ProductDBRecord(new Product(nextId(), "6 Milk Bricks Pack", 22 * 16 * 21, null), ProductFamily.DRINKS));
```
removing trailing space

use-cases/order-picking/src/main/java/org/acme/orderpicking/domain/Warehouse.java
```diff
@@ -0,0 +1,198 @@
...
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.acme.orderpicking.domain;
+
+import java.util.HashMap;
+import java.util.List;
```
yurloc:
```suggestion
```
unused import

use-cases/order-picking/src/main/java/org/acme/orderpicking/solver/OrderPickingConstraintProvider.java
```diff
@@ -0,0 +1,113 @@
...
+                //required buckets per order
+                .groupBy((trolley, order, orderTotalVolume) -> trolley,
+                        (trolley, order, orderTotalVolume) -> order,
+                        sum((trolley, order, orderTotalVolume) -> calculateOrderRequiredBuckets(orderTotalVolume, trolley.getBucketCapacity())))
+                //required buckets per trolley
+                .groupBy((trolley, order, orderTotalBuckets) -> trolley,
+                        sum((trolley, order, orderTotalBuckets) -> orderTotalBuckets))
+                //penalization if the trolley don't have enough buckets to hold the orders
+                .filter((trolley, orderTotalBuckets) -> trolley.getBucketCount() < orderTotalBuckets)
+                .penalize("Required number of buckets", HardSoftLongScore.ONE_HARD);
```
yurloc: @ge0ffrey would it be a good idea to consider the number of missing buckets in the penalty?

use-cases/order-picking/src/main/java/org/acme/orderpicking/util/LogUtil.java
```diff
@@ -0,0 +1,114 @@
...
+/**
+ * Helper class, not relevant.
+ */
+public class LogUtil {
+
+    private LogUtil() {
+    }
+
+    public static void log(Logger logger, OrderPickingSolution solution) {
+        if (solution == null) {
```
yurloc:
```suggestion
        if (!logger.isDebugEnabled()) {
            return;
        }
        if (solution == null) {
```
To avoid all of the hard work if nothing gets printed in the end.

yurloc (commented):
Reviewed the back end. Looks great. The Javadocs are awesome!

use-cases/order-picking/src/main/java/org/acme/orderpicking/bootstrap/DemoDataGenerator.java
yurloc: The reason it's lower is that the random selection produces duplicates that are not added but the loop counter still gets incremented.

wmedvede: I think it's ok. In the end what we want is to generate a random order. And the order is efectivelly generated randomly, as part of the items random generation we avoid adding duplicates for a matter of simplicity and not having to show thinigs like  Apples, Apples, Apples in the UI.

use-cases/order-picking/src/main/resources/META-INF/resources/warehouse-api.js
wmedvede: Originally I had organized the warehouse a bit different, then, after discussing with @ge0ffrey it was decided to have this (x,y) coordinates so my orignal modeling was changed to what we have now. But independently of this, I think It's ok, I my opinion people is acostumed to see an (x,y) system.

wmedvede:
@yurloc modifications where applied.

use-cases/order-picking/src/main/resources/META-INF/resources/warehouse-api.js
yurloc: I imagine that regular people, including non-technical, business people that might be watching demos of this, are familiar with the spreadsheet-like coordinate system. So yes, (x, y) but the origin should be in the top-left corner:
```
(0,0) ---- (1,0) ----→ x
  |
  |
(0,1)
  |
  |
  ↓
  y
```
because people start reading from the top. So in my opinion it makes more sense if the first row from the top is row 0 (or 1).

Also note that the change that I'm suggesting is a matter of presenting the data in the UI. It shouldn't affect the back end at all.

wmedvede: I see what you mean, and in fact this (0,0) origin at the top-left corner is what I had in my original design. But @ge0ffrey argued to change it into what we have now... with the opposit argument that you suggests, he said that "people are more accostumed to see what we have now"

yurloc: > he said that "people are more accostumed to see what we have now"

That surprises me.

Anyway, my suggestion is just a suggestion (doesn't block the PR) and I don't have any other arguments. Feel free to resolve this thread.

wmedvede: I think this could clarify the reasoning a bit more

SupermarketModel

build/quickstarts-showcase/src/main/resources/META-INF/resources/index.html
```diff
@@ -160,6 +160,29 @@ <h5>
                     </div>
                 </div>
             </div>
+            <div class="col mb-4">
+                <div class="card">
+                    <div class="card-header">
+                        <h5>
+                            Order picking
+                            <!-- TODO -->
```
ge0ffrey:
```suggestion
                            <!-- TODO adjust after merging https://github.com/kiegroup/optaplanner-website/pull/421/files -->
```

use-cases/order-picking/src/main/java/org/acme/orderpicking/solver/OrderPickingConstraintProvider.java
ge0ffrey: Yes, to avoid a score trap. Good catch!

ge0ffrey:
```suggestion
                .penalize("Required number of buckets", HardSoftLongScore.ONE_HARD,
                        (trolley, orderTotalBuckets) -> orderTotalBuckets - trolley.getBucketCount());
```

use-cases/order-picking/src/main/java/org/acme/orderpicking/util/LogUtil.java
```diff
@@ -0,0 +1,117 @@
...
+
+/**
+ * Helper class, not relevant.
+ */
+public class LogUtil {
+
+    private LogUtil() {
+    }
+
+    public static void log(Logger logger, OrderPickingSolution solution) {
```
ge0ffrey: Do we need to keep this method? It seems like something that was needed for debugging, but doesn't pull its weight in crispy, low-code quickstart. Wdyt?

use-cases/order-picking/src/main/java/org/acme/orderpicking/util/LogUtil.java
```diff
@@ -0,0 +1,117 @@
...
+        }
+        logger.debug("");
+        for (Trolley trolley : solution.getTrolleyList()) {
+            printAssignedOrdersPer(logger, trolley);
+        }
+        logger.debug("");
+        printUnAssignedOrders(logger, solution.getTrolleyStepList());
+    }
+
+    private static int getOrderRequiredVolume(Order order) {
```
ge0ffrey: I would not expect to find this method here when I look for it.
Can we move this method getOrderRequiredVolume on the Order class instead of the LogUtils class and just call it getRequiredVolume()?

use-cases/order-picking/src/main/java/org/acme/orderpicking/util/LogUtil.java
```diff
@@ -0,0 +1,117 @@
...
+        return order.getItems().stream()
+                .mapToInt(OrderItem::getVolume)
+                .sum();
+    }
+
+    private static void printAssignedOrdersPer(Logger logger, Trolley trolley) {
```
ge0ffrey: Do we need to keep this method now that we have a pretty UI?

use-cases/order-picking/src/main/java/org/acme/orderpicking/util/LogUtil.java
```diff
@@ -0,0 +1,117 @@
...
+                }
+            }
+        }
+        logger.debug("Orders not assigned in any Trolley, count: {}", orderList.size());
+        orderList.forEach(order -> {
+            logger.debug(" -> Order: {}", order.getId());
+        });
+    }
+
+    private static int getTotalVolumePerOrder(Order order, List<TrolleyStep> trolleySteps) {
```
ge0ffrey: Is this method still used if the print methods go away? Does it belong here?

use-cases/order-picking/src/main/java/org/acme/orderpicking/util/LogUtil.java
```diff
@@ -0,0 +1,117 @@
...
+        return trolleySteps.stream()
+                .filter(trolleyStep -> trolleyStep.getOrderItem().getOrder().getId().equals(order.getId()))
+                .mapToInt(trolleyStep -> trolleyStep.getOrderItem().getVolume())
+                .sum();
+    }
+
+    private static List<TrolleyStep> extractSteps(Trolley trolley) {
```
ge0ffrey: If this method is still used, can we move it to Trolley?

use-cases/order-picking/src/main/java/org/acme/orderpicking/util/LogUtil.java
```diff
@@ -0,0 +1,117 @@
...
+        TrolleyStep nextStep = trolley.getNextElement();
+        List<TrolleyStep> result = new ArrayList<>();
+        while (nextStep != null) {
+            result.add(nextStep);
+            nextStep = nextStep.getNextElement();
+        }
+        return result;
+    }
+}
```
ge0ffrey: If none of the methods in the class survive here, can we remove the LogUtils class?

use-cases/order-picking/src/main/resources/META-INF/resources/index.html
```diff
@@ -0,0 +1,118 @@
...
+
+<html lang="en">
+<head>
+    <meta content="text/html; charset=UTF-8" http-equiv="Content-Type">
+    <meta content="width=device-width, initial-scale=1" name="viewport">
+    <title>Order Picking - OptaPlanner Quarkus</title>
+    <link rel="stylesheet" href="/webjars/bootstrap/css/bootstrap.min.css"/>
+    <link rel="stylesheet" href="/webjars/font-awesome/css/all.css"/>
+    <link rel="stylesheet" href="/style.css"/>
+    <link rel="icon" href="favicon.gif" type="image/png">
```
ge0ffrey: If we add a favicon.gif for this quickstart, we need to do it for all quickstarts.

Not sure if that's worth it. It's very nice though.
Benefits: demo looks better and branding matters for marketing (that's why we have the optaplanner logo in every demo, which is crucial).
Downsides: one extra file for users to grok when going through the source code.

I guess it's worth it. But again, either we do it for all use case quickstarts or for none.

PS: I am also very surprised that a gif works. In the past, only an ico file worked for most browsers.

use-cases/order-picking/src/main/resources/META-INF/resources/tango-color.js
```diff
@@ -0,0 +1,67 @@
+/*
```
ge0ffrey, line 1: Good to put this in a seperate js. We'll need to do that for the other quickstarts too, but that can be in a separate PR.

use-cases/order-picking/src/main/resources/META-INF/resources/warehouse-api.js
ge0ffrey: I think an Excel approach is a good idea to make it easier to demo in front of an audience, to avoid the confusion as to what is the column and what is the row. In excel, the column is A, B, C etc and the row is 1, 2, 3 etc:

image

The downside is that behind the scenes, row 1 has index 0, row 5 index 4... Maybe we should start with A0, B0, etc?
To translate the column index 2 into C, do something like `String.valueOfCharacter('A' + columnIndex)`.

This can be in a separate PR - let's not drag this in the scope of this PR. Let's discuss first too.

use-cases/order-picking/src/main/resources/application.properties
```diff
@@ -0,0 +1,25 @@
...
+
+########################
+# OptaPlanner properties
+########################
+
+# The solver runs for 5 minutes. To run for 30 seconds use "30s" and for 2 hours use "2h".
+#quarkus.optaplanner.solver.termination.spent-limit=5m
+quarkus.optaplanner.solver.termination.spent-limit=120s
+
+%dev.quarkus.log.category."org.acme.orderpicking.rest.OrderPickingSolverResource".level=DEBUG
```
ge0ffrey: Is this log line worth its weight to explain to the user? I'd remove it to keep it simple. Was this for debugging while developing that particular piece, and can it be removed now?

On the other hand, can we add the other OptaPlanner properties commented out that we want users to be aware of that they exist?

```
# To change how many solvers to run in parallel
# optaplanner.solver-manager.parallel-solver-count=4
# To run increase CPU cores usage per solver
# quarkus.optaplanner.solver.move-thread-count=2

# To detect common bugs in your code
# quarkus.optaplanner.solver.environment-mode=FULL_ASSERT

# XML file for power tweaking, defaults to solverConfig.xml (directly under src/main/resources)
# quarkus.optaplanner.solver-config-xml=org/.../orderPickingSolverConfig.xml
```

use-cases/order-picking/src/main/resources/solverConfig.xml
```diff
@@ -0,0 +1,13 @@
...
+        xsi:schemaLocation="https://www.optaplanner.org/xsd/solver https://www.optaplanner.org/xsd/solver/solver.xsd">
+
+  <solutionClass>org.acme.orderpicking.domain.OrderPickingSolution</solutionClass>
+  <entityClass>org.acme.orderpicking.domain.TrolleyOrTrolleyStep</entityClass>
+  <entityClass>org.acme.orderpicking.domain.TrolleyStep</entityClass>
+
+  <scoreDirectorFactory>
+    <constraintProviderClass>org.acme.orderpicking.solver.OrderPickingConstraintProvider</constraintProviderClass>
+  </scoreDirectorFactory>
+</solver>
```
ge0ffrey: This file can be deleted, Quarkus will do all of this automatically. And if it doesn't, it's a bug.

use-cases/order-picking/src/test/java/org/acme/orderpicking/solver/OrderPickingConstraintProviderTest.java
```diff
@@ -0,0 +1,241 @@
...
+
+    @Inject
+    ConstraintVerifier<OrderPickingConstraintProvider, OrderPickingSolution> constraintVerifier;
+
+    private static Stream<Arguments> requiredNumberOfBucketsParams() {
+        return Stream.of(
+                Arguments.of(2, 5, 2, 5, 0),
+                Arguments.of(2, 4, 2, 5, 1),
+                Arguments.of(2, 5, 2, 4, 1),
+                Arguments.of(2, 4, 2, 4, 2));
```
ge0ffrey: Can we make this test easier to read by doing less magic? Just hard code the values like in the other ConstraintProviderTests.
=> no @ParameterizedTest

I'd argue only one of these lines needs to run, the others can be deleted. Allowing it to be inlined instead of parameterized.
Will that reduce test coverage? Test coverage is measured in number of lines touched, more specifically in number of branches touched. If all these numbers go through the same if-else branches, they do not increase test coverage.
If one of them does go through a different branch, so it's a corner case, then we do need it and it must be in a separate @Test method with a method name that signals the corner case reason, for maintenance and test failure reporting reasons.

use-cases/order-picking/src/test/java/org/acme/orderpicking/solver/OrderPickingConstraintProviderTest.java
```diff
@@ -0,0 +1,241 @@
...
+
+    private static Trolley mockTrolley(int bucketCount, int bucketCapacity, TrolleyStep... steps) {
+        Trolley trolley = new Trolley();
+        trolley.setBucketCapacity(bucketCapacity);
+        trolley.setBucketCount(bucketCount);
+        linkPathElements(trolley, steps);
+        return trolley;
+    }
+
+    private static List<TrolleyStep> trolleySteps(Trolley trolley) {
```
ge0ffrey, line 238: Duplicated in LogUtils. Would be nice to have on Trolley itself?

use-cases/order-picking/src/test/java/org/acme/orderpicking/solver/OrderPickingConstraintProviderTest.java
```diff
@@ -0,0 +1,241 @@
...
+                trolleySteps(order1Trolley3),
+                trolleySteps(order2Trolley1))
+                .flatMap(Collection::stream).toArray();
+
+        constraintVerifier.verifyThat(OrderPickingConstraintProvider::minimizeOrderSplitByTrolley)
+                .given(allSteps)
+                .penalizesBy(4 * 1000);
+    }
+
+    private static Order mockOrder(String id, OrderItem... items) {
```
ge0ffrey, line 198: because we don't use mockito or any other mocking lib, can we name it createOrder instead? Same for other mock methods. On the other hand, this is mocking. We can definitely leave this as is too.

ge0ffrey (approved):
Looks good to me. Please review the comments (fixing/rejecting as you see fit), especially the ones around the LogUtil class. We'll merge once you give the signal to merge.

use-cases/order-picking/src/main/resources/META-INF/resources/index.html
wmedvede: I guess I couldn't make the ico work, and thus, I changed it to the .png

use-cases/order-picking/src/main/java/org/acme/orderpicking/util/LogUtil.java
wmedvede: we can remove the logging altogether, np.,  also less code to maintain.

use-cases/order-picking/src/main/resources/solverConfig.xml
wmedvede: good point, yes, works fine without this file. Will be removed.

use-cases/order-picking/src/main/java/org/acme/orderpicking/util/LogUtil.java
wmedvede: removed

use-cases/order-picking/src/main/java/org/acme/orderpicking/util/LogUtil.java
wmedvede: removed

use-cases/order-picking/src/main/java/org/acme/orderpicking/util/LogUtil.java
wmedvede: removed

use-cases/order-picking/src/main/resources/META-INF/resources/index.html
wmedvede: removed

use-cases/order-picking/src/main/resources/META-INF/resources/warehouse-api.js
wmedvede: fixed, the structure was changed to the excel like format

use-cases/order-picking/src/main/java/org/acme/orderpicking/solver/OrderPickingConstraintProvider.java
wmedvede: suggestion added

build/quickstarts-showcase/src/main/resources/META-INF/resources/index.html
wmedvede: fixed

use-cases/order-picking/src/main/java/org/acme/orderpicking/util/LogUtil.java
wmedvede: has gone

use-cases/order-picking/src/main/java/org/acme/orderpicking/util/LogUtil.java
wmedvede: has gone

use-cases/order-picking/src/main/resources/solverConfig.xml
wmedvede: has gone

use-cases/order-picking/src/test/java/org/acme/orderpicking/solver/OrderPickingConstraintProviderTest.java
wmedvede: fixed

use-cases/order-picking/src/test/java/org/acme/orderpicking/solver/OrderPickingConstraintProviderTest.java
wmedvede: has gone

PR #196, opened 2021-10-07
Add a link to OptaPy for Python users in technology, so they can find it

ge0ffrey:


README.adoc
```diff
@@ -16,6 +16,7 @@ a|* link:use-cases/school-timetabling/README.adoc[Quarkus] (Java, Maven or Gradl
 * link:technology/java-spring-boot/README.adoc[Spring Boot] (Java, Maven or Gradle, Spring Boot, H2)
 * link:technology/java-activemq-quarkus/README.adoc[ActiveMQ] (Java, ActiveMQ, Maven, Quarkus)
 * link:technology/kotlin-quarkus/README.adoc[Kotlin] (Kotlin, Maven, Quarkus, H2)
+* link:technology/python/README.adoc[Python] (Python, Pip)
```
rsynek: It's definitely good to mention it here, but if we call it experimental on the website, the same we should call it here to keep our posture consistent across all places where we point to OptaPy.

ge0ffrey: Good point, even though this chain of links actually leads to the optapy.org page on github which mentions it's experimental. Anyway, can't hurt, let's do that.

ge0ffrey: Fixed :)

rsynek (approved):

PR #201, opened 2021-10-13
Readme commands order picking (parallel PR that didn't apply here yet)

ge0ffrey:
Like the Ghostbusters, the beams got crossed. A parallel PR did this for the other quickstarts after the order picking quickstart was PR'ed. This aligns that.


yurloc (approved):
LGTM, same as #197.

PR #212, opened 2021-10-26
Maintenance scheduling for road maintenance

ge0ffrey:
The "yellow" period is still missing.


use-cases/maintenance-scheduling/src/main/java/org/acme/maintenancescheduling/solver/MaintenanceScheduleConstraintProvider.java
```diff
@@ -32,103 +41,69 @@
...
-    public Constraint jobsMustStartAfterReadyTimeGrain(ConstraintFactory constraintFactory) {
-        return constraintFactory.fromUnfiltered(MaintenanceJobAssignment.class)
-                .filter(maintenanceJobAssignment -> maintenanceJobAssignment.getStartingTimeGrain() != null
-                        && maintenanceJobAssignment.getStartingTimeGrain().getGrainIndex() < maintenanceJobAssignment.getMaintenanceJob().getReadyTimeGrainIndex())
-                .penalizeConfigurable("Jobs must start after ready time grain",
-                        maintenanceJobAssignment -> maintenanceJobAssignment.getMaintenanceJob().getReadyTimeGrainIndex()
-                                - maintenanceJobAssignment.getStartingTimeGrain().getGrainIndex());
+    public Constraint crewConflict(ConstraintFactory constraintFactory) {
+        return constraintFactory
+                .fromUniquePair(Job.class,
```
triceo: Please convert the example to `forEach...()`, the `from...()` deprecation has already been merged.

use-cases/maintenance-scheduling/src/main/java/org/acme/maintenancescheduling/solver/MaintenanceScheduleConstraintProvider.java
```diff
@@ -32,103 +41,69 @@
...
 
-    public Constraint assignAllNonCriticalJobs(ConstraintFactory constraintFactory) {
-        return constraintFactory.fromUnfiltered(MaintenanceJobAssignment.class)
-                // Non critical maintenance jobs must be assigned a crew and start period
-                .filter(maintenanceJobAssignment -> !maintenanceJobAssignment.getMaintenanceJob().isCritical()
-                        && (maintenanceJobAssignment.getAssignedCrew() == null || maintenanceJobAssignment.getStartingTimeGrain() == null))
-                .penalizeConfigurable("Assign all non critical jobs");
-    }
-
-    public Constraint jobsShouldFinishBeforeSafetyMargin(ConstraintFactory constraintFactory) {
```
triceo, line 126: The safety margin was a key idea of the quickstart, and one of the main reasons why we were doing the quickstart in the first place.

It appears that the feature is now gone completely.

triceo (changes requested):
Needs API changes.
Also I suggest the removal of safety margin be reconsidered.

use-cases/maintenance-scheduling/src/main/java/org/acme/maintenancescheduling/solver/MaintenanceScheduleConstraintProvider.java
ge0ffrey: Yes, this is a regression, I 'll make sure this "yellow period" comes back.

use-cases/maintenance-scheduling/src/main/java/org/acme/maintenancescheduling/rest/MaintenanceScheduleResource.java
```diff
@@ -92,22 +86,22 @@ protected MaintenanceSchedule findById(Long id) {
         if (!SINGLETON_SCHEDULE_ID.equals(id)) {
             throw new IllegalStateException("There is no schedule with id (" + id + ").");
         }
-
+        // TODO get all working days (no weekend nor holidays) from database
+        LocalDate fromDate = LocalDate.now().with(TemporalAdjusters.next(DayOfWeek.MONDAY));
+        LocalDate toDate = fromDate.plusWeeks(10);
```
ge0ffrey: This is a bug. for small demo data it's 8 weeks, for large it's 16 weeks.

use-cases/maintenance-scheduling/src/main/java/org/acme/maintenancescheduling/solver/MaintenanceScheduleConstraintProvider.java
ge0ffrey: Fixed

use-cases/maintenance-scheduling/src/main/resources/META-INF/resources/index.html
```diff
@@ -63,22 +61,21 @@ <h1>Rental car maintenance schedule solver</h1>
...
         </div>
     </div>
-
     <div class="mb-4 tab-content" id="myTabContent">
-        <div class="tab-pane fade show active" id="employeeView" role="tabpanel" aria-labelledby="employeeViewTab">
+        <div class="tab-pane fade show active" id="byCrewPanel" role="tabpanel" aria-labelledby="byCrewPanelTab">
             <div id="employeeVisualization"></div>
         </div>
-        <div class="tab-pane fade" id="managerView" role="tabpanel" aria-labelledby="managerViewTab">
+        <div class="tab-pane fade" id="byJobPanel" role="tabpanel" aria-labelledby="byJobPanelTab">
```
triceo, line 78: The "Unassigned jobs" section below this. It looks weird when there is nothing in there.
The header is there, and then nothing. It's a bit confusing.
Consider adding a message, something like "All jobs have been assigned."

use-cases/maintenance-scheduling/src/main/java/org/acme/maintenancescheduling/solver/MaintenanceScheduleConstraintProvider.java
```diff
@@ -32,103 +41,93 @@
...
+        return constraintFactory
+                .forEachUniquePair(Job.class,
+                        overlapping(Job::getStartDate, Job::getEndDate),
+                        // TODO Use intersecting() when available https://issues.redhat.com/browse/PLANNER-2558
+                        filtering((job1, job2) -> !Collections.disjoint(
+                                job1.getMutuallyExclusiveTagSet(), job2.getMutuallyExclusiveTagSet())))
+                .penalizeLong("Mutually exclusive tag", HardSoftLongScore.ofSoft(1_000),
+                        (job1, job2) -> {
+                            Set<String> intersection = new HashSet<>(job1.getMutuallyExclusiveTagSet());
+                            intersection.retainAll(job2.getMutuallyExclusiveTagSet());
```
triceo: I think that `mutuallyExclusiveTagSet` is a horrible name.
The tags are not mutually exclusive between themselves - they are mutually exclusive with tags of other jobs.
It should be `tagSet` - the mutual exclusivity is brought by this constraint.

use-cases/maintenance-scheduling/src/main/java/org/acme/maintenancescheduling/solver/MaintenanceScheduleConstraintProvider.java
```diff
@@ -32,103 +41,93 @@
...
+    public Constraint mutuallyExclusiveTag(ConstraintFactory constraintFactory) {
+        // Avoid overlapping maintenance jobs with the same tag (for example road maintenance in the same area).
+        return constraintFactory
+                .forEachUniquePair(Job.class,
+                        overlapping(Job::getStartDate, Job::getEndDate),
+                        // TODO Use intersecting() when available https://issues.redhat.com/browse/PLANNER-2558
+                        filtering((job1, job2) -> !Collections.disjoint(
+                                job1.getMutuallyExclusiveTagSet(), job2.getMutuallyExclusiveTagSet())))
+                .penalizeLong("Mutually exclusive tag", HardSoftLongScore.ofSoft(1_000),
+                        (job1, job2) -> {
```
triceo, line 120: Seeing a consequence like this hurts my eyes. We've always been told: don't make RHS complex.
We may not have a better solution here, but I question if it's a pattern that we want to actively promote to people.
After all, you can completely ignore any filters and groupbys, and just do everything in a consequence.

triceo (approved):
Some comments, but no show stoppers.

use-cases/maintenance-scheduling/src/main/java/org/acme/maintenancescheduling/solver/MaintenanceScheduleConstraintProvider.java
ge0ffrey: This is to avoid a score trap: the number of common tags matters. And the number of overlapping days matter.

Of course, if we had that Matchers class, we could probably dumb this down to something like
`(job1, job2) -> Matchers.commonCount(Job::getTagSet) * Matchers.overlap(Job::getStartDate, Jog::getEndDate)`

ge0ffrey: Until then, this has to stay, in my opinion.

use-cases/maintenance-scheduling/src/main/resources/META-INF/resources/index.html
ge0ffrey: Fixed

triceo:
The new screenshot doesn't have the new message wrt. unassigned jobs.

use-cases/maintenance-scheduling/src/main/java/org/acme/maintenancescheduling/solver/MaintenanceScheduleConstraintProvider.java
ge0ffrey: Good point, resolved.

ge0ffrey:
> The new screenshot doesn't have the new message wrt. unassigned jobs.

Wierd, but it's in there now. Must have been in a later commit...

We'll want to merge this commits anyway, I think.

PR #219, opened 2021-11-10
PLANNER-2547: Make lines go around shelves

Christopher-Chianelli:
### JIRA

https://issues.redhat.com/browse/PLANNER-2547

image
image
image

<details>
<summary>
How to retest this PR or trigger a specific build:
</summary>

* for a <b>pull request build</b> please add comment: <b>Jenkins retest this</b>
* for a <b>full downstream build</b> please add the label `run_fdb`
* for a <b>specific pull request build</b> please add comment: <b>Jenkins (re)run [optaplanner-quickstarts] tests</b>
* for a <b>Quarkus LTS check</b> please add comment: <b>Jenkins run LTS</b>
* for a <b>specific Quarkus LTS check</b> please add comment: <b>Jenkins (re)run [optaplanner-quickstarts] LTS</b>
* for a <b>Native check</b> please add comment: <b>Jenkins run native</b>
* for a <b>specific Native LTS check</b> please add comment: <b>Jenkins (re)run [optaplanner-quickstarts] native</b>
</details>

### DIFF
--- a/use-cases/order-picking/src/main/resources/META-INF/resources/index.html
+++ b/use-cases/order-picking/src/main/resources/META-INF/resources/index.html
@@ -90,7 +90,7 @@
                 </div>
                 <div id="mapCanvasContainer" style="padding-top: 20px; padding-left: 15px;">
                     <!-- Canvas drawing by app.js -->
-                    <canvas height="1100px;" width="1200px;" id="warehouseCanvas"></canvas>
+                    <canvas height="1200px;" width="1300px;" id="warehouseCanvas"></canvas>
                 </div>
                 <div>
                     *Shelving dimension is of 2 meters width x 10 meters height.
--- a/use-cases/order-picking/src/main/resources/META-INF/resources/index.html
+++ b/use-cases/order-picking/src/main/resources/META-INF/resources/index.html
@@ -54,7 +54,7 @@
                     <a class="nav-link active" id="solutionTab" data-toggle="tab" href="#solutionTabDiv" role="tab" aria-controls="solutionPanel" aria-selected="true">Picking plan</a>
                 </li>
                 <li class="nav-item">
-                    <a class="nav-link" id="mapTab" data-toggle="tab" href="#mapTabDiv" role="tab" aria-controls="mapPanel" aria-selected="false">Map</a>
+                    <a class="nav-link" id="mapTab" data-toggle="tab" href="#mapTabDiv" role="tab" aria-controls="mapPanel" aria-selected="false" onclick="window.setTimeout(() => refreshSolution(), 500);">Map</a>
                 </li>
                 <li class="nav-item">
                     <a class="nav-link" id="unassignedEntitiesTab" data-toggle="tab" href="#unassignedEntitiesTabDiv" role="tab" aria-controls="unassignedEntitiesPanel" aria-selected="false">Unassigned</a>
@@ -88,9 +88,9 @@
                 <div id="mapActionsContainer">
                     <!-- Filled in by app.js -->
                 </div>
-                <div id="mapCanvasContainer" style="padding-top: 20px; padding-left: 15px;">
+                <div id="mapCanvasContainer" style="width: 75%; height: 75%; padding-top: 20px; padding-left: 15px;">
                     <!-- Canvas drawing by app.js -->
-                    <canvas height="1200px;" width="1300px;" id="warehouseCanvas"></canvas>
+                    <canvas id="warehouseCanvas" style="position: absolute"></canvas>
                 </div>
                 <div>
                     *Shelving dimension is of 2 meters width x 10 meters height.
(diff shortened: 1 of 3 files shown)


triceo:
Much better! One thing I noticed: in the second screenshot, the brown line goes needlessly all the way around the B2 isle.
Is that something that the solver came up with, or is that a bug in your algorithm?

triceo:
Also, the numbers of stops could use an improvement, I think. Here's what I propose:

The current situation is that the number of pickup shows up directly on top of the "dot" that signifies the pickup on the line.
How about instead we make the number go into one of the 4 corners of the dot? (north-west, north-east, south-east, south-west) That way, they would not overlap and the information would be easily readable.

Alternatively, we could use a sequence of numbers ("10, 11") in places where the numbers would otherwise be written over one another. That may be the better way to go here - especially if the numbers are color-coded to match the line.

ge0ffrey:
Beautiful!!!

- Can we make the shelves less thick so there is more room for the hallways?
Currently the shelves are twice as thick as the hallways. In reality, the shelves are about as thick as the hallways. But I 'd cheat and make the shelves half the size of the hallway, so we have more room to paint the lines.

- The shelves border can be less thick and "(A,2)" can be made "A2" to save space. So "(C,34)" would become "C34".
In theory, for 30 columns, we could end up with "AA" etc columns, like in Excel, but the UI must not be able to handle more than 26 columns.

- Does the UI separate the right lane from the left lane (opposite traffic) for clarity?
For example here, the pink trolley goes to up and then down to point 7:
Selection_916
In a 2 lane system, it would look like that:
image
Do we want to do that? Does it actually make it clearer?

use-cases/order-picking/src/main/resources/META-INF/resources/index.html
```diff
@@ -90,7 +90,7 @@ <h4><strong>It looks like the Order Picking quickstart has not been started...</
                 </div>
                 <div id="mapCanvasContainer" style="padding-top: 20px; padding-left: 15px;">
                     <!-- Canvas drawing by app.js -->
-                    <canvas height="1100px;" width="1200px;" id="warehouseCanvas"></canvas>
+                    <canvas height="1200px;" width="1300px;" id="warehouseCanvas"></canvas>
```
ge0ffrey: According to bootstrap 5
  https://getbootstrap.com/docs/5.0/layout/breakpoints/#available-breakpoints
these are typical window widths:

X-Small | None | <576px
-- | -- | --
Small | sm | ≥576px
Medium | md | ≥768px
Large | lg | ≥992px
Extra large | xl | ≥1200px
Extra extra large | xxl | ≥1400px


So a width of 1200px is a standard value, but a width of 1300px is not.
Any chance we can target at least "Extra large" or bigger screens again, not just "Extra extra large" screens?

ge0ffrey (approved):

ge0ffrey:
> Much better! One thing I noticed: in the second screenshot, the brown line goes needlessly all the way around the B2 isle. Is that something that the solver came up with, or is that a bug in your algorithm?

This, right?
image

ge0ffrey:
+1 to color-coded the numbers to match the line color

wmedvede:
Hi Guys, I haven't seen this PR.
Looks really good @Christopher-Chianelli !

A quick comment I detected while having a quick look at the changes working.
I believe there's a shift in the Steps numbers shown in the picking plan, and the numbers painted in the map.
This is what did, I've opened two different browser windows.
In window 1) I look at the picking plan for trolley1
In window 2) I look at the map for trolley1

So from the employee perspective, the warehouse location of step1 in the picking plan, must be the same location with numer1 in the map. I believe we have shift there. Have a look please.

wmedvede:
> > Much better! One thing I noticed: in the second screenshot, the brown line goes needlessly all the way around the B2 isle. Is that something that the solver came up with, or is that a bug in your algorithm?
> 
> This, right? image

I've noticed this too. But excellent work, looks really much better than the initial version!

Christopher-Chianelli:
image
image
image

triceo:
Chris, this looks really great now!
Would you mind incorporating the following, too? After that, I think we're good to merge.

> The current situation is that the number of pickup shows up directly on top of the "dot" that signifies the pickup on the line. How about instead we make the number go into one of the 4 corners of the dot? (north-west, north-east, south-east, south-west) That way, they would not overlap and the information would be easily readable.
> 
> Alternatively, we could use a sequence of numbers ("10, 11") in places where the numbers would otherwise be written over one another. That may be the better way to go here - especially if the numbers are color-coded to match the line.

Christopher-Chianelli:
Latest commits group order text:
image
image
image
The number are easy to read even if all the trolleys are selected (at the downside the offset can be large).

triceo (approved):
Cleaner, easier to read, nicer. Good job!

PR #223, opened 2021-11-15
 PLANNER-2559 Use bootstrap margin, padding, badge and card classes + remove trim()

ge0ffrey:
@rsynek I hope your fine with these changes, they went much further than originally intended on the call center use case. The goal was to standardize the use of bootstrap.

BEFORE

Selection_923

AFTER

Selection_922

### DIFF
--- a/technology/java-activemq-quarkus/client/src/main/resources/META-INF/resources/app.js
+++ b/technology/java-activemq-quarkus/client/src/main/resources/META-INF/resources/app.js
@@ -94,7 +94,7 @@ function refreshTimeTable() {
 
     $.each(timeTable.lessonList, (index, lesson) => {
       const color = pickColor(lesson.subject);
-      const lessonElementWithoutDelete = $(`<div class="card lesson" style="background-color: ${color}"/>`)
+      const lessonElementWithoutDelete = $(`<div class="card" style="background-color: ${color}"/>`)
         .append($(`<div class="card-body p-2"/>`)
           .append($(`<h5 class="card-title mb-1"/>`).text(lesson.subject))
           .append($(`<p class="card-text ml-2 mb-1"/>`)
--- a/technology/java-activemq-quarkus/client/src/main/resources/META-INF/resources/index.html
+++ b/technology/java-activemq-quarkus/client/src/main/resources/META-INF/resources/index.html
@@ -35,7 +35,7 @@
     <h1>High school time table solver</h1>
     <p>Generate the optimal schedule for your teachers and students.</p>
 
-    <div style="margin-bottom: .5rem">
+    <div class="mb-2">
         <button id="refreshButton" type="button" class="btn btn-secondary">
             <span class="fas fa-refresh"></span> Refresh
         </button>
@@ -61,7 +61,7 @@
             </ul>
         </div>
     </div>
-    <div class="tab-content" id="myTabContent">
+    <div class="tab-content mb-4" id="myTabContent">
         <div class="tab-pane fade show active" id="byRoomPanel" role="tabpanel" aria-labelledby="byRoomTab">
             <table class="table table-borderless table-striped timeTableSolution" id="timeTableByRoom">
                 <!-- Filled in by app.js -->
@@ -79,7 +79,7 @@
         </div>
     </div>
 
-    <h2 style="margin-top: 2rem">Unassigned lessons</h2>
+    <h2>Unassigned lessons</h2>
     <div id="unassignedLessons" class="card-columns"></div>
 </div>
 
--- a/technology/java-activemq-quarkus/common/src/main/java/org/acme/schooltimetabling/domain/Lesson.java
+++ b/technology/java-activemq-quarkus/common/src/main/java/org/acme/schooltimetabling/domain/Lesson.java
@@ -41,9 +41,9 @@ public class Lesson {
 
     public Lesson(Long id, String subject, String teacher, String studentGroup) {
         this.id = id;
-        this.subject = subject.trim();
-        this.teacher = teacher.trim();
-        this.studentGroup = studentGroup.trim();
+        this.subject = subject;
+        this.teacher = teacher;
+        this.studentGroup = studentGroup;
     }
 
     public Lesson(long id, String subject, String teacher, String studentGroup, Timeslot timeslot, Room room) {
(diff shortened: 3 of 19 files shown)


technology/java-activemq-quarkus/client/src/main/resources/META-INF/resources/app.js
```diff
@@ -94,7 +94,7 @@ function refreshTimeTable() {
 
     $.each(timeTable.lessonList, (index, lesson) => {
       const color = pickColor(lesson.subject);
-      const lessonElementWithoutDelete = $(`<div class="card lesson" style="background-color: ${color}"/>`)
+      const lessonElementWithoutDelete = $(`<div class="card" style="background-color: ${color}"/>`)
```
ge0ffrey, line 97: this lesson class was unused and didn't even exist

technology/java-activemq-quarkus/client/src/main/resources/META-INF/resources/index.html
```diff
@@ -35,7 +35,7 @@
     <h1>High school time table solver</h1>
     <p>Generate the optimal schedule for your teachers and students.</p>
 
-    <div style="margin-bottom: .5rem">
+    <div class="mb-2">
```
ge0ffrey, line 38: Like all other changes here: as we default on bootstrap, we use bootstraps m-, mb- (margin-bottom), mt-, p-, pb- (padding-bottom), pt-, etc.
Basic guidelines:
- mr-1 is between an icon and a text
- mb-2 is between two related pieces
- mb-4 is between two unrelated pieces

technology/java-activemq-quarkus/common/src/main/java/org/acme/schooltimetabling/domain/Lesson.java
```diff
@@ -41,9 +41,9 @@
 
     public Lesson(Long id, String subject, String teacher, String studentGroup) {
         this.id = id;
-        this.subject = subject.trim();
-        this.teacher = teacher.trim();
-        this.studentGroup = studentGroup.trim();
+        this.subject = subject;
+        this.teacher = teacher;
+        this.studentGroup = studentGroup;
```
ge0ffrey, line 46: Less is more.

use-cases/call-center/src/main/resources/META-INF/resources/app.js
```diff
@@ -33,36 +33,30 @@ function printCallTable(callCenterData) {
   const callTable = $('#callTable');
   callTable.children().remove();
   printHeader(callTable, 10);
-  const tableBody = $('<tbody>').appendTo(callTable);
+  const tableBody = $('<tbody/>').appendTo(callTable);
```
ge0ffrey, line 36: Generally, all $(...) elements must be closed. Otherwise the browser does it for you (with strange results).

use-cases/call-center/src/main/resources/META-INF/resources/app.js
```diff
@@ -33,36 +33,30 @@ function printCallTable(callCenterData) {
...
-  const agentCardRow = $(`<div class="row flex-nowrap">
-                <div class="col-4">
-                    <i class="fas fa-user-alt"></i>
-                </div>
-                <div class="col-8">
-                    <span style="font-size:1.2em">${agent.name}</span>
-                </div>
-        </div>`).appendTo(agentCardBody);
+  const agentCardBody = $('<div class="card-body p-2"/>')
+    .append($(`<h5 class="card-title mb-2"><i class="fas fa-user-alt mr-1"></i> ${agent.name}</h5>`));
```
ge0ffrey, line 54: I replaced the container-row and flex approach with a simple card.
This went much further than I originally intended, but I think it worked out fine with simpler code:

Selection_922

use-cases/call-center/src/main/resources/META-INF/resources/app.js
```diff
@@ -72,29 +66,24 @@ function printAgent(tableBody, agent) {
...
+    .append(callButtons)
+    .append($(`<h5 class="card-title mb-2"/>)`)
+      .append((call.pinned) ? $(`<i class="fas fa-phone-volume mr-1"></i>`) : $(`<i class="fas fa-phone mr-1"></i>`))
+      .append(call.phoneNumber))
+    .append();
+
+  printTimes(callCardBody, callButtons, call);
+  printSkills(callCardBody, call.requiredSkills);
+  callCard.append(callCardBody);
+  callCard.appendTo(callsTd);
```
ge0ffrey, line 83: same here, I replaced the container-row in a card approach with a normal card approach.

Fixes: on hover-over, the red "end call button" doesn't turn partially black any more
Side effects: the +1m button is now blue...

use-cases/call-center/src/main/resources/META-INF/resources/app.js
```diff
@@ -131,15 +118,11 @@ function formatDuration(duration) {
...
+  const skillRow = $('<div/>');
   container.append(skillRow);
   skills.forEach((skill) => {
     let color = skillToColorMap.get(skill);
-    skillRow.append($(`
-            <div class="col-xs-1 card" style="background-color:${color};margin:2px;padding:2px">
-                <span style="font-size:0.8em">${skill}</span>
-            </div>`)
-    );
+    skillRow.append($(`<span class="badge mr-1 mt-1" style="background-color:${color}">${skill}</span>`));
```
ge0ffrey, line 125: I replaced the card approach with a badge approach (used a lot these days for the other examples when they have skills).
Bug fixed: there is now padding between the skills divs
Side effect: their text is now bold

use-cases/call-center/src/main/resources/META-INF/resources/app.js
ge0ffrey: Just thinking, I can revert that side-effect, making it black again. wdyt?

rsynek: Yes, I think the black works better there; the blue button is not contrasty enough.

use-cases/call-center/src/main/resources/META-INF/resources/app.js
rsynek: Thanks, good to know. Anyway, this looks like a copy-paste mistake; we add unclosed elements in e.g. school timetabling too.

rsynek (approved):
Looks very good, thanks! Maybe just reverting the "+1m" to the black color would make it more contrasty.

PR #247, opened 2021-12-17
Upgrade Logback to 1.2.9

ge0ffrey:
https://github.com/kiegroup/optaplanner/pull/1733


Christopher-Chianelli (commented):
Is logback only used/defined in the "hello-world" quickstart?

ge0ffrey:
> Is logback only used/defined in the "hello-world" quickstart?

Directly yes. But we should also update Spring once they release a new version that uses a newer logback version.

PR #265, opened 2022-01-13
nightly GHA job

Ginxo:
### JIRA
https://issues.redhat.com/browse/PLANNER-2604

Can you please guys fill the gaps? from `Set failure message`, `Set success message` and `Send a stream message` steps?

Once we have it clear I will create `ZULIP_API_KEY` secret before merging.
I set two different event types:

- `schedule` to be run based on cron expression `0 0 * * *` (every night)(for default branch, stable in this case)
- `workflow_dispatch` to manually run it (choosing the branch)

You can check execution tests here https://github.com/Ginxo/optaplanner-quickstarts/actions/workflows/nightly.yml

<details>
<summary>
How to retest this PR or trigger a specific build:
</summary>

* for a <b>pull request build</b> please add comment: <b>Jenkins retest this</b>
* for a <b>full downstream build</b> please add the label `run_fdb`
* for a <b>specific pull request build</b> please add comment: <b>Jenkins (re)run [optaplanner-quickstarts] tests</b>
* for a <b>Quarkus LTS check</b> please add comment: <b>Jenkins run LTS</b>
* for a <b>specific Quarkus LTS check</b> please add comment: <b>Jenkins (re)run [optaplanner-quickstarts] LTS</b>
* for a <b>Native check</b> please add comment: <b>Jenkins run native</b>
* for a <b>specific Native LTS check</b> please add comment: <b>Jenkins (re)run [optaplanner-quickstarts] native</b>
</details>

PR #266, opened 2022-01-13
[development] nightly GHA job

Ginxo:
https://github.com/kiegroup/optaplanner-quickstarts/pull/265 backported to development just to keep stable/development aligned

### DIFF
--- a/.github/workflows/nightly.yml
+++ b/.github/workflows/nightly.yml
@@ -1,0 +1,56 @@
+
+name: Nightly
+
+on:
+  schedule:
+    - cron: "0 0 * * *"
+  workflow_dispatch:
+
+jobs:
+  nightly:
+    concurrency:
+      group: nightly
+      cancel-in-progress: true
+    strategy:
+      matrix:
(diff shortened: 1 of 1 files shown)


rsynek:
I am thinking if we actually need it for the `development` branch. The main goal was to make sure that what users get has not been broken during the release. Users get the `stable` (default) branch.

Ginxo:
@rsynek we don't need it, and this file won't affect on anything apart from being aligned with the things from `stable` branch. Only default branch flows are taken into consideration by GHA (apart from the ones from PR's source branch, but this is not the case)

rsynek:
Actually, we need this file, because otherwise, the `development` branch would override it on the `stable` branch during the release (`stable` is reset to the last commit of the release branch).

rsynek (approved):

Ginxo:
> Actually, we need this file, because otherwise, the `development` branch would override it on the `stable` branch during the release (`stable` is reset to the last commit of the release branch).

That's what I meant, sorry if I didn't explain myself well. Thanks

PR #292, opened 2022-02-15
PLANNER-2657: Add native OpenShift Configuration for school timetabling

Christopher-Chianelli:
To deploy on OpenShift, first login to your OpenShift account using `oc login ...`, then execute the following command in `use-cases/school-timetabling`: 

`mvn clean package -Dopenshift -Dquarkus.kubernetes-client.trust-certs=true -Dquarkus.kubernetes.deploy=true -DskipTests -Dquarkus.native.container-build=true -Dquarkus.native.container-runtime=podman`

`quarkus.kubernetes-client.trust-certs=true ` is if you are using a local instance/a test instance using self-signed certificates. ` -Dquarkus.native.container-build=true -Dquarkus.native.container-runtime=podman` is to force a container build so the native image will run on OpenShift (will get linking issues otherwise). `-Dquarkus.kubernetes.deploy=true` is to deploy automatically after the build (can apply config in `target/kubernetes/openshift.yml` instead)

### JIRA

https://issues.redhat.com/browse/PLANNER-2657

### Referenced pull requests

<details>
<summary>
How to retest this PR or trigger a specific build:
</summary>

* for a <b>pull request build</b> please add comment: <b>Jenkins retest this</b>
* for a <b>full downstream build</b> please add the label `run_fdb`
* for a <b>specific pull request build</b> please add comment: <b>Jenkins (re)run [optaplanner-quickstarts] tests</b>
* for a <b>Quarkus LTS check</b> please add comment: <b>Jenkins run LTS</b>
* for a <b>specific Quarkus LTS check</b> please add comment: <b>Jenkins (re)run [optaplanner-quickstarts] LTS</b>
* for a <b>Native check</b> please add comment: <b>Jenkins run native</b>
* for a <b>specific Native LTS check</b> please add comment: <b>Jenkins (re)run [optaplanner-quickstarts] native</b>
</details>

### DIFF
--- a/use-cases/school-timetabling/pom.xml
+++ b/use-cases/school-timetabling/pom.xml
@@ -193,7 +193,7 @@
       </properties>
     </profile>
     <profile>
-      <id>openshift</id>
+      <id>openshift</id> <!-- Optional for use in OpenShift. -->
       <activation>
         <property>
           <name>openshift</name>
--- a/use-cases/school-timetabling/pom.xml
+++ b/use-cases/school-timetabling/pom.xml
@@ -192,6 +192,28 @@
         <quarkus.profile>native</quarkus.profile>
       </properties>
     </profile>
+    <profile>
+      <id>openshift</id>
+      <activation>
+        <property>
+          <name>openshift</name>
+        </property>
+      </activation>
+      <properties>
+        <quarkus.package.type>native</quarkus.package.type>
+        <quarkus.profile>openshift</quarkus.profile>
+      </properties>
+      <dependencies>
+        <dependency>
+          <groupId>io.quarkus</groupId>
+          <artifactId>quarkus-openshift</artifactId>
+        </dependency>
+        <dependency>
+          <groupId>io.quarkus</groupId>
+          <artifactId>quarkus-jdbc-postgresql</artifactId>
+        </dependency>
+      </dependencies>
+    </profile>
   </profiles>
 
   <repositories>
--- a/use-cases/school-timetabling/pom.xml
+++ b/use-cases/school-timetabling/pom.xml
@@ -193,15 +193,15 @@
       </properties>
     </profile>
     <profile>
-      <id>openshift</id> <!-- Optional for use in OpenShift. -->
+      <id>openshift-native</id> <!-- Optional for use in OpenShift. -->
       <activation>
         <property>
-          <name>openshift</name>
+          <name>openshift-native</name>
         </property>
       </activation>
       <properties>
         <quarkus.package.type>native</quarkus.package.type>
-        <quarkus.profile>openshift</quarkus.profile>
+        <quarkus.profile>openshift-native</quarkus.profile>
       </properties>
       <dependencies>
         <dependency>
--- a/use-cases/school-timetabling/src/main/kubernetes/openshift.yml
+++ b/use-cases/school-timetabling/src/main/kubernetes/openshift.yml
@@ -1,0 +1,142 @@
+apiVersion: v1
+items:
+  - apiVersion: v1
+    kind: Secret
+    metadata:
+      annotations:
+        template.openshift.io/expose-database_name: '{.data[''database-name'']}'
+        template.openshift.io/expose-password: '{.data[''database-password'']}'
+        template.openshift.io/expose-username: '{.data[''database-user'']}'
+      labels:
+        app: postgresql
+        app.kubernetes.io/component: postgresql
+        app.kubernetes.io/instance: postgresql
+        template: postgresql-ephemeral-template
+      name: postgresql
+    stringData:
+      database-name: app
+      database-password: app
+      database-user: app
+  - apiVersion: v1
+    kind: Service
+    metadata:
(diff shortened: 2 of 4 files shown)


Christopher-Chianelli:
(this is different from the S2I work, which gets the Java version up and running; this work is to get the native version of the app up and running)

use-cases/school-timetabling/pom.xml
```diff
@@ -192,6 +192,28 @@
         <quarkus.profile>native</quarkus.profile>
       </properties>
     </profile>
+    <profile>
+      <id>openshift</id>
```
triceo:
```suggestion
      <id>openshift</id> <!-- Optional for use in OpenShift. -->
```

use-cases/school-timetabling/src/main/resources/application.properties
```diff
@@ -53,3 +53,16 @@ quarkus.hibernate-orm.database.generation=drop-and-create
 
 # In pom.xml, the "native" maven profile triggers the "native" quarkus profile.
 %native.quarkus.datasource.jdbc.url=jdbc:h2:tcp://localhost/mem:school-timetabling
+
+########################
+# OpenShift overrides
```
triceo:
```suggestion
# Optional overrides for use in OpenShift
```

use-cases/school-timetabling/src/main/resources/application.properties
```diff
@@ -53,3 +53,16 @@ quarkus.hibernate-orm.database.generation=drop-and-create
...
+########################
+%openshift.quarkus.openshift.name=school-timetabling
+%openshift.quarkus.openshift.route.expose=true
+%openshift.quarkus.datasource.db-kind=postgresql
+%openshift.quarkus.datasource.jdbc.url=jdbc:postgresql://postgresql/app
+
+# Do not store username and password in application.properties for a production application;
+# Use .env or environment variables instead
```
triceo:
```suggestion
# Note: Do not store username and password in application.properties for a production application;
#       Use .env or environment variables instead
```

triceo (commented):
The description of this PR is good, and it belongs in the quickstart's `README`.

The Kubernetes config is a lot of code, and I have to admit I am uncomfortable shipping this without any sort of regression testing. What are our options here?

Christopher-Chianelli:
In regards to regression testing, we would need to run it against an OpenShift instance. I am guessing maybe a Jenkins pipeline? Start new OpenShift instance, log in, run the maven command, check services are expose, get routes, test the solve route (or maybe add a healthcheck route)?

triceo:
If my memory serves, the Kogito guys already have something like that figured out. `kogito-apps`, I think. Maybe we find our inspiration there?

Christopher-Chianelli:
I left out the instructions for insecure certificates in the README since we direct them to the developer sandbox, which should have signed certificates.

Christopher-Chianelli:
Looking at `kogito-app`, they use test containers (https://www.testcontainers.org/); not so sure if that will work in this use case, since we are testing a maven build (and as such, a JUnit test won't be able to test it)

triceo:
I checked a few random Quarkus quickstarts and it seems that they have no OpenShift coverage either. Unfortunate. I guess we're just going to have to merge as is.

@yurloc Thoughts?

use-cases/school-timetabling/src/main/kubernetes/openshift.yml
```diff
@@ -0,0 +1,142 @@
+apiVersion: v1
```
triceo, line 1: While browsing the Quarkus quickstarts, I stumbled upon this:
https://github.com/quarkusio/quarkus-quickstarts/tree/main/getting-started-knative

It appears that their convention is to store Kubernetes configs in `/k8s`. Let's do what they do for consistency.

Christopher-Chianelli: that.... goes against quarkus documentation for the `quarkus-kubernetes` extension: https://quarkus.io/guides/deploying-to-kubernetes#using-existing-resources

Christopher-Chianelli:
(Side note: don't merge yet; I need to check the old method still work (IIRC, S2I builds activate the openshift profile, and the profile I add was "openshift", which causes it to use Postgres instead of in memory H2 (In memory H2 does not work for native, hence the kubernetes config)).

use-cases/school-timetabling/src/main/kubernetes/openshift.yml
triceo: Strange. Ok, let's go with what the documentation says. Maybe we report an issue on that quickstart, so that they fix it?

yurloc:
I'm afraid it's not possible to have this covered by an automated test without having a permanently available OpenShift cluster. @dupliaka does some OpenShift testing for OptaWebs but that's only for the product I guess.

Christopher-Chianelli:
Can confirm using the `openshift` profile broke S2I; using `openshift-native` instead fixed it and now both S2I and native work.

PR #638, opened 2025-09-26
[issue-639] Disable scheduled nigthly as not used

mareknovotny:
Removing scheduled nightly and kept only manual triggering for a case until all problems with used actions are sorted out in https://github.com/apache/incubator-kie-optaplanner-quickstarts/issues/639


mareknovotny:
fyi @pibizza @baldimir 
Could you guys help me to turn off daily notification it failed due actions are not approved?

